PR #2, opened 2016-04-04
Took out a lot more files that do not belong in this repo

adamjohnwright:
@mgrbyte  This is the beginnings of removing the parts of the code that are not necessary for the REST API between Datomic and the website.
## Testing
- I ran the website within the REPL to make sure all dependencies are met
- I went to datomic.wormbase.org to make sure data was sill able to be displayed on the website without error
## Notable
- Both the expression and phenotype widgets were not being populated (error status) on the website. I am hoping that this is because the data in the database is incomplete. If that isn't true it could also be caused by something external to this tool. Will have to figure out with @sibyl229.
# Todo
- I will need to change out the pseudoace namespace with an external module that @mgrbyte will be creating as a part of his re-factoring efforts. 
- Remove further unused scripts in the wb folder
- Change the folder structure (once all unused code is removed)
- Create other REST endpoints


mgrbyte:
Looks OK to me @a8wright.

PR #5, opened 2016-06-15
Feature/adam add tests

adamjohnwright:
In this merge there are three major things. 

1) I have added in the core functionality of midje tests. I hope to be substancially adding to these as I am developing

2)  I reduced the number of dependencies by removing the querying / authentication code.

3) this code has been upgraded to work with pseudoace 0.4.6

Testing:

I have been running and compiling the code with lein repl. Everything compiles without error. I have also curled a couple URLs to make sure the data is able to be pulled correctly from the API. 

I am hoping this provides a more solid foundation to move forward with making sure the json being provided by the interface is matching that produced by the AceDB REST API. 

@mgrbyte I think at this point it is a good idea for you to compile the code with lein repl but don't need to test too much further than that. I will be working on fixing any bugs in the next round of changes. Also I am not using the formatting from emacs at this point. I will be incorporating emacs into my development over the summer. Feel free to fix formatting if it takes minimal effort.


mgrbyte:
@a8wright Generally, looks good to me.

I got errors running `lein repl` due to it not being able to find the datomic.pro library.

How do you set up your environment so that this works without that?

This didn't do the trick, so I added the following to project.clj (this line is the same in the pseudoace jar):

``` clojure
  :repositories {"my.datomic.com" {:url "https://my.datomic.com/repo"
                                   :creds :gpg}}
```

I was then able to run `lein repl` cleanly.

PR #7, opened 2016-07-25
Feature/adam complete gene interface

adamjohnwright:
## Description

In this feature branch I removed several unnecessary parts of the code, changed the structure of the code with regards to the database connection and added several queries to make the JSON schema match that provided by ACeDB. 

The database connection now uses "mount" (https://github.com/tolitius/mount). Matt and I were deciding between using "Components" (https://github.com/stuartsierra/components to handle the database connections. The reason I decided on using mount is because it doesn't force us to use components throughout the program. 

Before creating the first release I will be commenting out all but a select few trusted URLs. I then will be releasing URL's on an individual basis.
### 0.0.1

This release will contain several URL's that have including but not limited to history and external_links.  
### 0.0.2

This release will contain several URL's that are more complex to check but are currently likely correct.


src/datomic_rest_api/rest/gene.clj
```diff
@@ -1235,15 +1237,72 @@
        (fn [ec-id]
          (let [ec (entity db ec-id)]
            {:expression_cluster (datomic-rest-api.rest.object/pack-obj "expression-cluster" ec)
-            :description        (:expression-cluster/description ec)}))))
+            :description      (str (:expression-cluster/description ec))}))))
      :description
      "expression cluster data"}))
 
+(defn- anatomic-expression-patterns [gene]
```
mgrbyte, line 1294: 2 space indent in block below for consistency with other code

src/datomic_rest_api/get_handler.clj
```diff
@@ -24,37 +24,37 @@
 
 (defroutes get-routes
   (GET "/rest/widget/gene/:id/overview" {params :params}
-       (gene/overview (db con) (:id params)))
```
mgrbyte, line 27: 2 space indent for this routes block?

PR #8, opened 2016-07-25
Feature/prep for deployment

adamjohnwright:
In this branch I have fixed up how the datomic-pro dependency is configured in the project.clj. This change now makes it so that the user has to specify to use the profiles datomic-pro and ddb if they are wanting to use the tool with the DynamoDB version of database. 

I have also commented out all the URLs but the external_links URL. This URL is by far the least complex and should serve as a great starting point for testing the API in production. 
### 0.0.1

This release will contain several URL's that have including but not limited to external_links.
### 0.0.2

This release will add the history and the phenotype URLs
### 0.0.3

This release will contain several URL's that are more complex to check but are currently likely correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,14 +16,14 @@
 lein deploy clojars
 ```
 
-###Create war file
+###Create jar file
 ```
 lein with-profile +datomic-pro,+ddb uberjar
 ```
 
 ####Test deploying the jar
 
-To see inside the war
+To see inside the jar
 ```
 jar tvf ./target/<jar-name>.jar
 ```
--- a/README.md
+++ b/README.md
@@ -3,10 +3,32 @@
 - Provide interface to datomic database to WormBase website. 
 
 ##Setting environment variables
-    export TRACE_PORT=8130
     export TRACE_DB="datomic:ddb://us-east-1/wormbase/WS254"
 
-## starting server
+##Starting server in development
 
-    lein repl
-    >  (use 'web.core)
+    lein with-profile +datomic-pro,+ddb ring server-headless 8130
+
+##Deploying to production
+
+###Deploy to Clojars (to be used by other clojar projects)
+```
+lein deploy clojars
+```
+
+###Create war file
+```
+lein with-profile +datomic-pro,+ddb uberjar
+```
+
+####Test deploying the jar
+
+To see inside the war
+```
+jar tvf ./target/<jar-name>.jar
+```
+
+To do a test deploy
+```
+sudo java -server -jar <jar-name>.jar
+```
--- a/project.clj
+++ b/project.clj
@@ -3,10 +3,8 @@
   :min-lein-version "2.0.0"
   :dependencies 
   [[org.clojure/clojure "1.8.0"]
-   [com.datomic/datomic-pro "0.9.5385"
-    :exclusions [joda-time]] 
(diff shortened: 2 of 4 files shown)


src/datomic_rest_api/rest/gene.clj
```diff
@@ -551,22 +560,126 @@
    :description
    "phenotype based on interaction"}))
         
-   
+(defn drives-overexpression [gene]
+  (let [db (d/entity-db gene)]
```
mgrbyte, line 564: ~~I think this is an error. Should be:
`(d/entity db gene)`, which means you'll need to pass `db` into the function.~~
This is intended. I have missed the fact that `entity-db` is part of the datomic api.

src/datomic_rest_api/rest/gene.clj
```diff
@@ -551,22 +560,126 @@
...
-
-   :phenotype
-   {:data
-    {:Phenotype              (phenotype-table (d/entity-db gene) (:db/id gene) false)
-     :Phenotype_not_observed (phenotype-table (d/entity-db gene) (:db/id gene) true)}
-    :description
-    "The phenotype summary of the gene"}
+   :drives_overexpression
+    {:data
+      {:Phenotype (phenotype-overexpression (d/entity-db gene) (:db/id gene))}
```
mgrbyte, line 672: ~~Same as above `d/entity-db` -> `d/entity`~~
This is intended. I have missed the fact that `entity-db` is part of the datomic api.

README.md
```diff
@@ -3,10 +3,32 @@
...
+    lein with-profile +datomic-pro,+ddb ring server-headless 8130
+
+##Deploying to production
+
+###Deploy to Clojars (to be used by other clojar projects)
+```
+lein deploy clojars
+```
+
+###Create war file
```
mgrbyte: Says "war" but creates "jar"

README.md
```diff
@@ -3,10 +3,32 @@
...
+####Test deploying the jar
+
+To see inside the war
+```
+jar tvf ./target/<jar-name>.jar
+```
+
+To do a test deploy
+```
+sudo java -server -jar <jar-name>.jar
```
mgrbyte, line 33: Command is fine, but we probably want to specify heap size too (i.e `XMX=<N>g`)

project.clj
```diff
@@ -31,6 +29,9 @@
   :plugins [[lein-cljsbuild "1.1.3"]
             [lein-pprint "1.1.1"]
             [lein-ring "0.9.7"]]
+  :datomic-pro {:dependencies [[com.datomic/datomic-pro "0.9.5385"
```
mgrbyte: This should be removed as it's specified under `:profiles` below

README.md
adamjohnwright: I have made the change "war" to "jar"

src/datomic_rest_api/rest/gene.clj
adamjohnwright, line 672: I will fix this for the next release. Still working on this section of the code.

src/datomic_rest_api/rest/gene.clj
adamjohnwright, line 564: I will fix this for the next release. Still working on this section of the code.

project.clj
adamjohnwright: I have removed it

PR #10, opened 2016-08-15
Overview URI

adamjohnwright:
This pull request is for changes to the Overview URI. 

This involved changing a few queries to make it so the json structure matches that produced by the AcEDB API. 

This also involved creating a couple queries. The main one being the for the "clone" section. 

I have queried the database with several URL's. Everything matched with these URLs. I have looked through the code in-depth and I don't see any major points in the logic of the code that could result in problems with edge case situations. 

For each of the URI's that are release to Dev will be looked at by curators to check to make sure there are no edge cases that weren't accounted for. 

The next URL has not been picked yet. I will be determining this based on the URI that is relatively simple and is unlikely to have any edge cases where there could be errors within the code. (least amount of risk).

### DIFF
--- a/src/datomic_rest_api/rest/gene.clj
+++ b/src/datomic_rest_api/rest/gene.clj
@@ -82,7 +82,7 @@
   {:data
    (let [db   (d/entity-db gene)
          cds  (:gene/corresponding-cds gene)
-         data {:defined_by_mutation (not (empty? (:variation.gene/_gene gene)))
+         data {:defined_by_mutation (if (not (empty? (:variation.gene/_gene gene))) 1 0)
                :type (cond
                       ;; This is pretty-much the reverse order of the Perl code
                       ;; because we never over-write anything
@@ -102,7 +102,7 @@
                       :default
                       (some #(transcript-type (:gene.corresponding-transcript/transcript %))
                             (:gene/corresponding-transcript gene)))
-               :associated_sequence (not (empty? cds))
+               :associated_sequence (if (not (empty? cds)) 1 0)
                :confirmed (if (q '[:find ?conf-gene .
                                    :in $ ?conf-gene
                                    :where [?conf-gene :gene/corresponding-cds ?gc]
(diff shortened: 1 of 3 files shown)


src/datomic_rest_api/rest/gene.clj
```diff
@@ -82,7 +82,7 @@
   {:data
    (let [db   (d/entity-db gene)
          cds  (:gene/corresponding-cds gene)
-         data {:defined_by_mutation (not (empty? (:variation.gene/_gene gene)))
+         data {:defined_by_mutation (if (not (empty? (:variation.gene/_gene gene))) 1 0)
```
mgrbyte: Does web api expect 0 or 1 in this field rather than the actual object?

mgrbyte, line 85: Does the json not expect to be `true` or `false` rather than `1` or `0` ? (assume not, otherwise you'd not have coded it this way - it's just surprising!)

src/datomic_rest_api/rest/gene.clj
```diff
@@ -102,7 +102,7 @@
                       :default
                       (some #(transcript-type (:gene.corresponding-transcript/transcript %))
                             (:gene/corresponding-transcript gene)))
-               :associated_sequence (not (empty? cds))
+               :associated_sequence (if (not (empty? cds)) 1 0)
```
mgrbyte, line 105: Same as comment above regarding truth-y json values

src/datomic_rest_api/rest/gene.clj
adamjohnwright, line 85: You are correct. The reason why would be because Perl does not have boolean variable types. This is something we can change if we start querying the API from the browser (JavaScript).

mgrbyte, line 85: :+1:

src/datomic_rest_api/rest/object.clj
```diff
@@ -245,12 +248,11 @@
...
-   :Date_last_updated
+   :Date_last_updated 
    (if-let [date (:evidence/date-last-updated holder)]     
-       { :id (str (.format (java.text.SimpleDateFormat. "yyyy-M-d") date))
-         :label (str (.format (java.text.SimpleDateFormat. "yyyy-M-d") date))
-         :class "text"})
-       
+       [{ :id (str (.format (java.text.SimpleDateFormat. "yyyy-M-dd") date))
+         :label (str (.format (java.text.SimpleDateFormat. "yyyy-M-dd") date))
+         :class "text"}])
```
mgrbyte, line 255: This is fine, but could probably use more clojure-like idioms by using the [clj-time.format](https://github.com/clj-time/clj-time) library, rather than using java.text (our projects usually include this library as a dependency, so no extra cost)

PR #27, opened 2016-10-18
Ensure login before push to ECR.

### LINKED ISSUE #26: Command to push the docker image to ECR should auto-login
Like the `make eb-local` command does.

Marked as bug because the README is written w/out stating need to login to ECR.

mgrbyte:
Fixes #26

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ jar tvf ./target/<jar-name>.jar
 
 To test:
 ```bash
-java -server -jar <jar-name>.jar
+java -jar <jar-name>.jar
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,7 +30,7 @@ jar tvf ./target/<jar-name>.jar
 
 To test:
 ```bash
-java -server -jar <jar-name>.jar
+java -jar <jar-name>.jar
```
mgrbyte, line 33: `-server` is the default and not required (even a no-op on the HotSpot JVM)

PR #28, opened 2016-11-04
(NOT A PULL) printing :variation.predicted-cds component using seq on the map

sibyl229:
Hi @mgrbyte 

This is NOT a pull request, just the code you asked for (getting the content in variation.predicted-cds component). Please feel free to delete the branch whenever you are done. 

My text editor removes trailing whitespaces, so the diff might show more changes than the genuine ones. You might need to do a diff ignoring whitespace changes:
https://github.com/WormBase/datomic-to-catalyst/compare/temp/predicted-cds-holder?expand=1&w=0

### DIFF
--- a/src/datomic_rest_api/get_handler.clj
+++ b/src/datomic_rest_api/get_handler.clj
@@ -52,7 +52,7 @@
 ;;     (GET "/rest/widget/gene/:id/reagents" {params :params}
 ;;         (gene/reagents db (:id params) (str "rest/widget/gene/" (:id params) "/reagents"))) ;; looks correct; needs sort to confirm
 ;;     (GET "/rest/widget/gene/:id/gene_ontology" {params :params}
-;;         (gene/gene-ontology db (:id params) (str "rest/widget/gene/" (:id params) "/gene_ontology"))) ;; substancially same structure. not producing the same results 
+;;         (gene/gene-ontology db (:id params) (str "rest/widget/gene/" (:id params) "/gene_ontology"))) ;; substancially same structure. not producing the same results
 ;;     (GET "/rest/widget/gene/:id/expression" {params :params}
 ;;         (gene/expression db (:id params) (str "rest/widget/gene/" (:id params) "/expression"))) ;; This one is predominantly done but needs a little checking and sequence data
 ;;     (GET "/rest/widget/gene/:id/homology" {params :params}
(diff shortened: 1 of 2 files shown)


mgrbyte:
@sibyl229 

I'm going to work on a more full-fledged solution to the problem we discsused on the hangout,
but in the meantime, I've worked up a function that should help (provided below).
It returns a map with the keys `:ref-schema` and/or `:component-schema`, which will point to schema names (symbols) that you see in the generated-schema EDN (e.g schema256.edn).

You'd use it like `(resolve-xref-info db ident)`
Where:
- `db` is a datomic database handle (as returned by `datomic.api/db`)
- `ident` is usually an attribute (e.g `:variation/predicted-cds`, `:gene/remark`, `:gene/reference` etc)

Here's an example of using it from a `cider` session within emacs:

```
 datomic-rest-api.get-handler> (resolve-xref-info DB :variation/predicted-cds)
{:component-schema molecular-change, :ref-schema variation.predicted-cds}
```

You can drop the following function definition somewhere into your codebase/repl,
then use it within the repl/code:

``` lisp
(defn resolve-xref-info
  "Resolve object references to an attribute `ident` given a `db` handle.

  Returned map may contain the following keys:
     `:ref-schema` - symbol of the referenced schema
     `:component-schema` - symbol of the component schema if `ident`
                           is a component attribute.
  "
  [db ident]
  (if-let [attr (d/entity db ident)]
    (if (:db/isComponent attr)
      {:component-schema (-> (:pace/use-ns attr)
                             (first)
                             (symbol))
       :ref-schema (symbol (str (namespace ident)
                                "."
                                (name ident)))}
      (if-let [ref (:pace/obj-ref attr)]
        {:ref-schema (-> ref
                         (namespace)
                         (symbol))}
        ))
    (throw (Exception. (str ident
                            " is not a valid schema attribute")))))
```

Hope that helps!

sibyl229:
Hi @mgrbyte thank you so much! I just tried it, the solution works great. (Though please try not to work on weekends =P) I'm gonna use this function in the repl for now. I wonder if this function will be in the pseudoace module that I can import, or shall I put this function definition somewhere in the datomic-to-catalyst repo. Any thoughts?

mgrbyte:
@sibyl229 I think it belongs in the pseudoace module. I'm not against you putting it datomic-to-catalyst temporarily, or better yet in a [dev](http://thinkrelevance.com/blog/2013/06/04/clojure-workflow-reloaded) `source-path`  e.g `./dev/user.clj` (assuming you envision the need for this function to be for debugging/development purposes only?)

sibyl229:
Thank you @mgrbyte !  I tried the source-paths you suggested. It's really helpful. I will create a new pull request in a bit. 

I learned that by setting `:repl-options {:init-ns user}` in the project.clj, we can change the initial namespace for the repl. I didn't actually do it, as it changes the current behaviour. Just mentioning here in case we want to change the behaviour in the future.  I'm just closing this pull request, as it's not needed any more.

PR #30, opened 2016-11-11
Genetics widget and tests

sibyl229:
Hi @mgrbyte @a8wright this pull request has a bit more changes. They are on the API endpoint for the genetics widget, and some test cases for sanity checks on the data coming out. There is no rush getting it merged though, please take your time. 

I made a few commits before realizing my text editor removed trailing spaces on lines I'm not editing. So for the genuine changes, please look at the diff that ignores whitespace changes here:
https://github.com/WormBase/datomic-to-catalyst/compare/feature/add-genetics?expand=1&w=1

Thanks!

### DIFF
--- a/src/datomic_rest_api/get_handler.clj
+++ b/src/datomic_rest_api/get_handler.clj
@@ -31,6 +31,7 @@
                       <li>/rest/widget/gene/:id/overview</li>
                       <li>/rest/widget/gene/:id/history</li>
                       <li>/rest/widget/gene/:id/mapping_data</li>
+                      <li>/rest/widget/gene/:id/genetics</li>
                     </ul>
                   </html>")
      (GET "/rest/widget/gene/:id/overview" {params :params}
@@ -61,8 +62,8 @@
 ;;         (gene/sequences db (:id params) (str "rest/widget/gene/" (:id params) "/sequences")))
 ;;     (GET "/rest/widget/gene/:id/feature" {params :params}
 ;;         (gene/features db (:id params) (str "rest/widget/gene/" (:id params) "/feature"))) ;; has a few values missing for gbrowse - not sure if needed or possible to get out of datomic
-;;     (GET "/rest/widget/gene/:id/genetics" {params :params}
-;;         (gene/genetics db (:id params) (str "rest/widget/gene/" (:id params) "/genetics"))) ;; looks good need to find rearrangement that exists. Also need to test if works, when does not exist, with Perl template
+     (GET "/rest/widget/gene/:id/genetics" {params :params}
+         (gene/genetics db (:id params) (str "rest/widget/gene/" (:id params) "/genetics")))
      (GET "/rest/widget/gene/:id/external_links" {params :params}
          (gene/external-links db (:id params) (str "rest/widget/gene/" (:id params) "/external_links")))))
 
--- a/src/datomic_rest_api/get_handler.clj
+++ b/src/datomic_rest_api/get_handler.clj
@@ -18,15 +18,7 @@
             [datomic-rest-api.rest.locatable-api :refer (feature-api)]))
 
 
-(defn rest-field [field-name field-fn]
-  (fn [db class id]
-    (let [wbid-field (str class "/id")]
-      (-> {:name id
-           :class class
-           (keyword field-name) (field-fn (d/entity db [(keyword wbid-field) id]))}
-          (json/generate-string)
-          (ring.util.response/response)
-          (ring.util.response/content-type "application/json")))))
+(declare handle-field-get)
 
 (defn app-routes [db]
    (routes
@@ -77,11 +69,9 @@
          (gene/genetics db (:id params) (str "rest/widget/gene/" (:id params) "/genetics")))
      (GET "/rest/widget/gene/:id/external_links" {params :params}
           (gene/external-links db (:id params) (str "rest/widget/gene/" (:id params) "/external_links")))
-     (GET "/rest/field/:class/:id/:field-name" [class id field-name]
-          (let [field-namespace (symbol (str "datomic-rest-api.rest." class))
-                field-fn (ns-resolve field-namespace (symbol (str/replace field-name "_" "-")))
-                restified-field-fn (rest-field field-name field-fn)]
-            (restified-field-fn db class id)))))
+     (GET "/rest/field/:class/:id/:field" [class id field]
+          (handle-field-get db class id field))))
+
 
 (defn init []
   (print "Making Connection\n")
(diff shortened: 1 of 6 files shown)


mgrbyte:
@sibyl229 Hi, sorry for taking so long to get back to you on this.

Would you mind making a branch of develop that includes all whitespace only changes, then re-basing this branch on-top of that?

(Reviewing this PR using the github "review changes" feature is not possible from the link you provided in the description)

One thing that sticks out is that is seems like the tests run against either the product or staging database? (or at least the database pointed to by the TRACE_DB environment variable)  - is that correct?
If so, I think we need to look into generating test databases / mocking the database state or review what/how we are testing in datomic-to-catalyst. (It's great that you've added tests though! :)

sibyl229:
@Thanks @mgrbyte for reviewing!

Yes, I will clean up the white spaces then rebase. And let you know when I'm done.

Those tests need a database to run. They don't modify any data in there, only read from it. Is this your concern? What they do is sanity check some output out of the rest api, and see if they are what we expect (both in terms of structure and actual content). This isn't an ideal solution, but it allows us to quickly check if data model changes break any critical part of the web display.

mgrbyte:
@sibyl229 Thanks! the rebase will help a lot :+1: 

Concerning the tests:
I think sanity checking the output is a fine.
In the short-medium term, no data is being modified, although in the long term this may no longer hold true.
When the tests "hit" the api, this consumes (presumably dynamodb) resources, thus there is a (albeit small) cost.

More pragmatic concerns/things to think about (not criticisms):
 * the tests are not quick** (for me) to run outside of the ec2 network
 * TRACE_DB environment variable needs setting appropriately to some external database URI (which one [do you use] ?)
  * Are the tests targeted to the current database or the upcoming one (e.g WS256 or WS257) ?
    If the data/display logic changes between releases, then tests will break against one and pass against the other depending on the state of the code; I imagine we always want the tests to run against "upcoming".
 * If we were wanting to automate testing (ala continuous integration) how would we do that?

** I think they would be if I were to go and checkout the repo on an ec2 host and run them from there.
The performance/speed issue is network latency outside of the ec2 region/network.
 
IMHO, tests (esp. unit tests) should:
  * be self-contained (not require an external resource to run)
  * run quickly (less than a minute or so)

I think what we need is a way to generate a small subset of the database suitable for testing,
and store that in the project's  `resources/dev` directory (or the like) and load this during tests instead of the pointing to the real database.
 
In the absence of such a dataset, I think the current approach is better than no tests - but I do think we _should prioritize_ creation of such a test database.

Thinking aloud: 
   One approach would be to migrate a database from selected subsets of the most recently migrated database.
   e.g: from a set of .ace files, use pseudoace to produce a set of edn files, and migrate then at run time in the tests with `use-fixture :once`.

sibyl229:
Hi @mgrbyte Thanks for putting lots thought into testing! Some of the things are harder to answer than others, so I will comment on what I already know and have thought about.

> the tests are not quick** (for me) to run outside of the ec2 network

I understand speed of tests is important. I think there should be ways to run subset of tests, hopefully based on profile loaded or some other setting. I will take a look into it. For now, maybe not run the tests locally? 

>TRACE_DB environment variable needs setting appropriately to some external database URI (which one [do you use] ?)

My trace-db setting targets WS256 currently. I think it makes sense to have the tests observe the trace-db value,which is used by the rest api app. Is there any concern about this during the build process?

>If the data/display logic changes between releases, then tests will break against one and pass against the other depending on the state of the code; I imagine we always want the tests to run against "upcoming".

Right, test made this way will be vulnerable against **data change between releases**, especially **deletion or modification** of existing value. It can be robust against **addition**, by not replying the position or order of objects. On the other hand, maybe those tests should be excluded from the tests that run regularly "between releases" (such as restarting a production server), because when it fails, it's always a false positive. What do you think?  (I'm not sure what you meant by display logic change, could you please clarify?) 

>If we were wanting to automate testing (ala continuous integration) how would we do that?

I haven't organized my thoughts on this. I feel based on your comment that there should be basic tests (as integration tests) that focus on obtaining connection to other services and dependencies, ie datomic-to-catalyst needs to confirm a connection to a datomic database. There should also be more thorough (and data specific) tests built into an automated process for development, like running local development or deploy to staging environment. 

> tests (esp. unit tests) should be self-contained (not require an external resource to run)

I agree with your point here. The challenge is in developing and maintaining a self contained data set that:
- exactly reflect the state of the model ( => this makes generating mock data **from scrach** difficult)
- the tags are sufficiently populated ( => this makes generating mock data by slicing the actual database difficult, because the tags tend to be sparsely populated in the actual database, the tag you want to test for might not be included in the slice of data you retrieved).

I feel that, for this approach to succeed, there probably needs to be automated construction of the mock database from schema file. I haven't got any idea how to do that, or even this is the right way. I'm happy to hear your thoughts on this.

And @mgrbyte , could you let me know what fix is needed to get this PR merged? Thanks!

mgrbyte:
@sibyl229 Thanks for taking the time here; I think we are on the same page for all the points,
and with regard to testing. I think you, @a8wright and I should discuss this  further going forward.

I'll conclude the review this PR this morning. thanks!

src/datomic_rest_api/get_handler.clj
```diff
@@ -82,3 +89,40 @@
       (integer? p) p
       (string? p)  (parse-int p)
        :default default)))
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; internal functions and helper ;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn- wrap-response [data]
```
mgrbyte, line 98: Function body should use 2 space indentation

src/datomic_rest_api/rest/gene.clj
```diff
@@ -2191,34 +2226,56 @@
...
                       [?var :variation/gene ?vh]
-                      (not [?var :variation/allele _])]
+                      (not [?var :variation/allele _])
+                      (not [?var :variation/phenotype _])]
              db (:db/id gene))
           (map #(process-variation (entity db %))))
-     :description "polymorphisms and natural variations contained in the strain"}))
+     :description "polymorphisms and natural variations currently with no associated phenotype"}))
+
+(defn- alleles-count [gene]
```
mgrbyte, line 2235: I think the duplication here could be factored out. not critical for merge, perhaps we can pair on this later.

test/datomic_rest_api/rest/gene_test.clj
```diff
@@ -4,7 +4,113 @@
...
+(defn get-gene [id]
+  (d/entity (d/db datomic-conn) [:gene/id id]))
+
+; This is a regular test function, which is to be wrapped using my-test-fixture
+(deftest test-alleles-and-polymorphisms
+  (let [reference-allele (#'gene/reference-allele (get-gene "WBGene00006759"))
+        alleles (#'gene/alleles (get-gene "WBGene00006759"))
+        alleles-other (#'gene/alleles-other (get-gene "WBGene00006759"))
+        polymorphisms (#'gene/polymorphisms (get-gene "WBGene00006759"))
+        allele1 (first (filter #(= "WBVar00248722"
```
mgrbyte: Possibly pull out the duplicated expression to  get data for variation here.

mgrbyte (changes requested):
Generally, looks good! :+1: 
In gene.clj, I think it'd be better form to use an `:as` alias for datomic-rest-api.rest.object, e.g:
```clojure
(require '[datomic-rest-api.rest.object :as rest-api-obj])
```
which would make for shorter lines and increase readability.

src/datomic_rest_api/rest/gene.clj
sibyl229: Pair on this would be nice. There is actually a deeper issue hinted here with large tables where fetch all the data would take a long time. The counts is an indicator that there are other alleles (alleles_other field) and polymorphisms, that user can fetch that's not initially fetched with the widget. But how it's currently done is kinda a hack. I'd like to change it soon.

sibyl229:
Hi @mgrbyte I figured out the error about namespace `datomic-rest-api.utils.db` when using test-refresh. I mistakenly included a `require` of that namespace in dev/user.clj, where that namespace isn't defined. The tests are passing for me now, and the code duplication in test has been fixed.

I will do the alias fix you suggest in a separate config. 

Hope large PR like this doesn't happen in the future. Thanks a lot for reviewing :)

mgrbyte (approved):

PR #35, opened 2016-12-01
Feature/refactor controller

sibyl229:
Hi @a8wright and @mgrbyte hope you guys have fun at Clojure conj. When you have time (no rush), here is a PR for you guys :-)

**tl;dr This is a refactor of the `get-handler.clj` and `def-rest-widget`, that is fully compatible with everything else that has been implemented.**

A couple things this PR does:
- make `def-rest-widget` more generic: you can import it from rest/core.clj and use it anywhere
- make `def-rest-widget` macro trivial: 
     - the complexity is split into helper functions: widget-adaptor and json-response
     - at the moment, `def-rest-widget` is roughly synonymous to a `defn` (you can in fact use `defn` to declare widgets, not recommended though for forward compatibility)
- remove hard coded routes in `get-handler.clj`
- include a whitelist for widgets and fields that is checked in the controller/handler

Things NOT resolved by this PR:
- answer question like: can you tell me the fields available in the Gene Overview widget, without instantiate it? This might require either change to how `def-rest-widget` is being **called**, and/or its implementation. 

Thank you!

### DIFF
--- a/src/datomic_rest_api/get_handler.clj
+++ b/src/datomic_rest_api/get_handler.clj
@@ -41,8 +41,8 @@
                   </html>")
      (GET "/rest/widget/:class/:id/:widget" [class id widget :as request]
           (handle-widget-get db class id widget request))
-     (GET "/rest/field/:class/:id/:field" [class id field]
-          (handle-field-get db class id field))))
+     (GET "/rest/field/:class/:id/:field" [class id field :as request]
+          (handle-field-get db class id field request))))
 
 
 (defn init []
@@ -97,7 +97,8 @@
     (let [wrapped-field-fn (wrap-field field-fn)
           data (wrapped-field-fn db class id)]
       (-> {:name id
-           :class class}
+           :class class
+           :url (:uri request)}
           (assoc (keyword field-name) data)
           (wrap-response)))
     (-> {:message "field not exist or not available to public"}
--- a/src/datomic_rest_api/get_handler.clj
+++ b/src/datomic_rest_api/get_handler.clj
@@ -12,6 +12,7 @@
             [environ.core :refer (env)]
             [mount.core :as mount]
             [datomic-rest-api.utils.db :refer [datomic-conn]]
+            [datomic-rest-api.rest.core :refer [wrap-field wrap-widget]]
             [datomic-rest-api.rest.gene :as gene]
             [datomic-rest-api.rest.interactions :refer (get-interactions get-interaction-details)]
             [datomic-rest-api.rest.references :refer (get-references)]
@@ -19,6 +20,7 @@
 
 
 (declare handle-field-get)
+(declare handle-widget-get)
 
 (defn app-routes [db]
    (routes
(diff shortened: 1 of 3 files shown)


src/datomic_rest_api/rest/core.clj
```diff
@@ -0,0 +1,21 @@
+(ns datomic-rest-api.rest.core
+  (:require [datomic.api :as d :refer (db q touch entity)]
```
mgrbyte: Don't need the :refer part here, since not using the names directly:
```clojure
'[require datomic.api :as d]
```

mgrbyte (changes requested):
Generally, I _really_ like what you've done here, it's awesome! :+1: 

Other than one nit pick below, this looks excellent.

Great work @sibyl229!

src/datomic_rest_api/get_handler.clj
```diff
@@ -12,13 +12,15 @@
             [environ.core :refer (env)]
             [mount.core :as mount]
             [datomic-rest-api.utils.db :refer [datomic-conn]]
+            [datomic-rest-api.rest.core :refer [field-adaptor widget-adaptor]]
```
mgrbyte: Another minor nit-pick - personally, I prefer to use list notation for `:refer` names, e.g:
```clojure
[datomic-rest-api.utils.db :refer (datomic-conn)]
``` 
This is just for disambiguation; Don't mind either way (braces or brackets) but it would be good to keep the style consistent (esp. within a module, ideally within a project/org)

src/datomic_rest_api/get_handler.clj
```diff
@@ -37,40 +39,10 @@
...
-;;         (gene/features db (:id params) (str "rest/widget/gene/" (:id params) "/feature"))) ;; has a few values missing for gbrowse - not sure if needed or possible to get out of datomic
-     (GET "/rest/widget/gene/:id/genetics" {params :params}
-         (gene/genetics db (:id params) (str "rest/widget/gene/" (:id params) "/genetics")))
-     (GET "/rest/widget/gene/:id/external_links" {params :params}
-          (gene/external-links db (:id params) (str "rest/widget/gene/" (:id params) "/external_links")))
-     (GET "/rest/field/:class/:id/:field" [class id field]
-          (handle-field-get db class id field))))
+     (GET "/rest/widget/:class/:id/:widget" [class id widget :as request]
+          (handle-widget-get db class id widget request))
+     (GET "/rest/field/:class/:id/:field" [class id field :as request]
```
mgrbyte: Minor nit-pick number 3.... ;p

Would prefer to `ident` rather than `class` here as the name for two reasons:
1) class (I assume) refers to the ACeDB concept ; here the variable refers to the datomic version of the name (i think!), in which cases "entity-name" or "ident" would be a better name imho
2) `class` is a clojure builtin, and this "shadows" it.

sibyl229: Hi @mgrbyte Thanks a lot for looking at it closely :) These are all very fair points. I will correct them. 

The value taken on here (in the url) is something like like `gene`, `person`.  I wonder maybe "schema-name" or "n(ame)s(pace)-name" would be more fitting? I don't exactly understand the concept of `ident`, they seem to be readable alias for specific entity IDs, and is often used when declare attribute in a schema, I guess attribute definitions are entities too?

mgrbyte: Either `schema-name` or `ns-name` works great for me, and in fact is more technically correct.
fwiw, `ident` is the datomic concept of "identity" of an entity. 

e.g: if you have a gene entity/entity map, then:
 ```clojure
`(= (:db/ident gene) :gene/id)
```
holds true.
<<ns-name/schema-name>>/id is equvilant to the "name"/"id" of an ACeDB object.

[1]: http://docs.datomic.com/identity.html

Comment (mgrbyte) [reply src/datomic_rest_api/get_handler.clj]: `:<thing>/id` is something that was chosen to present the primary ACeDB object identifier (e.g WBGene00000001 for genes) in the schema translation; where as `:db/ident` is the attribute that points to the value that represents the attribute identity in the schema (e.g `:gene/id`, `:sequence/id`)
It happens that `:<thing>/id` is nearly always chosen as the value for `:db/ident`, but that may not always/necessarily be true (i.e is subject to change depending on the entity in question).

**Comment by sibyl229** [comment]
Thank you @mgrbyte . That clarified things for me. 
Your suggested fixes have all been done now. 

There is a minor thing I just noticed. Github syntax highlighting is confused about the list form in `:refer` and thought it's a function call o.O
```clojure
[datomic.api :as d :refer (db history q touch entity)]
```
```clojure
[datomic.api :as d :refer [db history q touch entity]]
```

mgrbyte (approved):

PR #37, opened 2016-12-06
Feature/phenotype

adamjohnwright:
I have created and tested the Phenotype URI. This URL to my knowledge creates identical JSON as Catalyst does.

In order to make sure everything is correct I have gone through a substantial amount of the code with @mgrbyte  and will be going over this code with @sibyl229.

Thanks everybody for the help!

@tharris This is fairly close we just have to go through the proper testing procedure with the pull request and with getting a curator or two to check over the results.

### DIFF
--- a/src/datomic_rest_api/get_handler.clj
+++ b/src/datomic_rest_api/get_handler.clj
@@ -31,7 +31,7 @@
                        <li><a href=\"./rest/field/\">/rest/field/</a></li>
                     <ul>
                   </html>")
-     (GET "/rest/fiedl/" [] "<html>
+     (GET "/rest/field/" [] "<html>
                             <ul>
                               <li>/rest/field/gene/:id/alleles-other</li>
                               <li>/rest/field/gene/:id/polymorphism</li>
--- a/src/datomic_rest_api/get_handler.clj
+++ b/src/datomic_rest_api/get_handler.clj
@@ -44,7 +44,7 @@
                       <li>/rest/widget/gene/:id/history</li>
                       <li>/rest/widget/gene/:id/mapping_data</li>
                       <li>/rest/widget/gene/:id/genetics</li>
-                      <li>/rest/widget/gene/:id/phenotype</li>
+                      <li>/rest/widget/gene/:id/phenotypes</li>
                     </ul>
                   </html>")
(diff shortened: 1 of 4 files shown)


src/datomic_rest_api/rest/gene.clj
```diff
@@ -533,12 +533,13 @@
...
                             (keys rnai-phenos)))]
     (->>
-      (for [pid phenos :let [pheno (entity db pid)]]
-        (let [pcs (get-pato-combinations db pid rnai-phenos var-phenos not?)]
-          (if (nil? pcs)
-            (phenotype-table-entity db pheno nil nil pid var-phenos rnai-phenos not?)
-            (for [[pato-key entity] pcs]
-              (phenotype-table-entity db pheno pato-key entity pid var-phenos rnai-phenos not?)))))
+      (flatten
+        (for [pid phenos :let [pheno (entity db pid)]]
```
mgrbyte: I think the style convention for `:let` in forms brings it onto its own line e.g:
```clojure
(for [x xs
      :let [n-xs (count xs)]]
    (do-something x))
```

mgrbyte (commented):
I think this formatting of `where` clauses is fine,
however I prefer the following:
```clojure
:where
 [clause-1]
 ....
 [clause-n]
```
since this helps avoids long lines.

(General comment on c4da5f0)

src/datomic_rest_api/get_handler.clj
```diff
@@ -28,15 +28,23 @@
                     <h5>Widgets</h5>
                     <ul>
                        <li><a href=\"./rest/widget/\">/rest/widget/</a></li>
+                       <li><a href=\"./rest/field/\">/rest/field/</a></li>
                     <ul>
                   </html>")
+     (GET "/rest/fiedl/" [] "<html>
```
mgrbyte: I assume there's a typo here, should be: 
(GET "/rest/field" ...

src/datomic_rest_api/rest/gene.clj
adamjohnwright: I have pushed the fix for this

src/datomic_rest_api/get_handler.clj
adamjohnwright: This was a type - fixed

mgrbyte (approved):
:+1:

Issue #44, opened 2017-01-11
Problem compiling with mount db file (db.clj)


adamjohnwright:
##Problem
This error occurs when you switch branches or move a file (db.clj), presumably with a macro, to a new location.

##Output

lein with-profile +datomic-pro,+ddb ring server-headless 8130
Picked up _JAVA_OPTIONS: -Xmx4G
Picked up _JAVA_OPTIONS: -Xmx4G
Compiling datomic-rest-api.get-handler  
^[[A^[[APicked up _JAVA_OPTIONS: -Xmx4G 
^[[AException in thread "main" java.lang.NoClassDefFoundError: mount/core/DerefableState, compiling:(/tmp/form-init3934453147777406180.clj:1:73)
        at clojure.lang.Compiler.load(Compiler.java:7391)

adamjohnwright:
##Fix 

1) Run lein to get the error
2) touch -m datomic_to_catalyst/utils/db.clj
3) Should be fixed - rerun lein

PR #46, opened 2017-01-13
remove problematic ahead-of-time compilation

sibyl229:
Hi @a8wright @mgrbyte @azurebrd  I think avoiding Ahead-of-time compilation might be a fix for the mount problem #44, along with the problem that the server doesn't always update when `rest/fields/gene.clj` is modified.

**Please run `lein clean` before starting the server for the _first time_ with this change.** This hopefully remove the product of previous compilation. It seems to work for me consistently. Please let me know whether this solution works for you guys. Thanks!

I don't have a full explanation for the fix. I don't understand the details about lein's compilation process myself. But based on the two issues #44 and the reloading problem, it appears that we get different compilation results based on timestamps of the source file. So the experiment has been to simplify compilation settings in the `project.clj` file, in particular avoid things that might cause a "stateful" compilation.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -29,8 +29,9 @@
   :plugins [[lein-cljsbuild "1.1.3"]
             [lein-pprint "1.1.1"]
             [lein-ring "0.9.7"]]
-  :main datomic-rest-api.get-handler
+  :main ^:skip-aot datomic-rest-api.get-handler
   :aot [datomic-rest-api.get-handler]
+  :uberjar {:aot :all}
   :ring {:handler datomic-rest-api.get-handler/app
          :host "0.0.0.0"
          :init datomic-rest-api.get-handler/init}
--- a/project.clj
+++ b/project.clj
@@ -30,7 +30,6 @@
             [lein-pprint "1.1.1"]
             [lein-ring "0.9.7"]]
   :main datomic-rest-api.get-handler
-  :aot [datomic-rest-api.get-handler]
   :ring {:handler datomic-rest-api.get-handler/app
          :host "0.0.0.0"
          :init datomic-rest-api.get-handler/init}


adamjohnwright:
Thanks @sibyl229  I am going to try this out in the  morning

mgrbyte:
@sibyl229 nice find!
More detail on the aot compilation process [here](https://clojure.org/reference/compilation).

project.clj
```diff
@@ -30,7 +30,6 @@
             [lein-pprint "1.1.1"]
             [lein-ring "0.9.7"]]
   :main datomic-rest-api.get-handler
```
mgrbyte: Think we still need to specify [skip-aot](mount/src/mount/) for the `:main` namespace, e.g:

```
:main ^:skip-aot  datomic-to-catalyst.get-handler
```

Additionally, I think we want to preserve aot for uberjar builds.
This can be done by adding:
```
:uberjar {:aot :all}
```
to the `:profiles` map.

adamjohnwright: @mgrbyte this makes sense to me. Our configuration was consistently working with the aot. It was just the runtime that was having troubles.

mgrbyte: @a8wright yep. I've submitted a PR upstream to the mount repo that should fix the underlying issue (as per ref above)

sibyl229:
Thanks @mgrbyte @a8wright for catching the uberjar issue! The 
`clojure
{:uberjar {:aot :all}}` option you set should be sufficient to resolve the uberjar issue (without requiring `:aot [datomic-rest-api.get-handler]`, based on [project.clj doc on :aot](https://github.com/technomancy/leiningen/blob/master/sample.project.clj#L255-L257). 

It appears that the :skip-aot doesn't take effect for me, and I had to remove the `:aot [datomic-rest-api.get-handler]` in d95b31ac32bdf15879e2ba5a77759ff925eb5cf1 to avoid the mount issue. [project.clj doc on :main hinted about this](https://github.com/technomancy/leiningen/blob/master/sample.project.clj#L206-L207) @a8wright and I tested creating the uberjar after the change, and it appears to work. Just letting you know here about the change. Please let me know if this breaks anything.

PR #49, opened 2017-01-22
Feature/add feature

adamjohnwright:
This pull request contains the "sequence features" widget url: curl -H content-type:application/json http://localhost:8130/rest/widget/gene/WBGene00000149/feature

This is the first URL with a connection to the Sequence mySQL database. I believe we now have the right approach with connecting to the database and, since the last restart of the mySQL server, everything seems stable.

We are using HugSQL. It works really well. I organized our folder structure and ns system based on the example app from HugSQL. I think that it s what we want. Although I am open to suggestions. 

I have a system for creating the "right" window for viewing the gene features in GBrowse. This system is not identical to the one in the Perl code. Although this is on purpose - the Perl code is quite convoluted. the answers the range for viewing comes out to be extremely similar (within a 50 bases usually) to the ones produced by the Perl code. I think these difference are caused by logic in the CPAN module Bio::SeqFeature::Store modules. I don't think they are necessary. 

##Testing
I have also tested the rest of the parts of the URL. I did this by finding examples of each of the sub fields. Some were quite sparsely populated.
I have also tested the code by sampling random urls (about ten or so) and compared them to that which comes out of the the Perl code. They are identical.

@sibyl229 I would like going through the how this appears on the website. See if you see anything that is different bettween the two sytstems. 
@matt feel free to test it out and see if it works with your IDE setup... Cider. 

I want to make sure this doesn't fail. I am not sure how to test what happens when the connection comes and goes. I am pretty certain that it will not be a problem. I just want to see if either of you can see any issues or know of a way of testing the system based on the stability of the connection to the database.


src/datomic_rest_api/rest/fields/gene.clj
```diff
@@ -2085,81 +2087,128 @@
...
+                              (double (- padded-stop padded-start)))))
+                       padded-stop)
+        id (str (:seqname segment) ":" browser-start ".." browser-stop) ]
+    {:class "genomic_location" ;; To populate this correctly we will need sequence data
+     :id id
+     :label id
+     :pos_string id
+     :taxonomy (if-let [class (:gene/species gene)]
+                 (if-let [[_ genus species] (re-matches #"^(.*)\s(.*)$" (:species/id class))]
+                   (clojure.string/lower-case (clojure.string/join [(first genus) "_" species]))))
```
mgrbyte: use `str/lower-case`, `str/join`

src/datomic_rest_api/rest/fields/gene.clj
```diff
@@ -2085,81 +2087,128 @@
...
+(defn- segment-to-position [gene segment gbrowse]
+  (let [start (if (< (:start segment) (:end segment))
+                (:start segment)
+                (:end segment))
+        stop (if (> (:start segment) (:end segment))
+               (:start segment)
+               (:end segment))
+        padded-start (- start 2000)
+        padded-stop (+ stop 2000)
+        browser-start (if (= gbrowse true)
```
mgrbyte: consider (perhaps?) factoring out browser-start & browser-stop calculations:
e.g
```clojure
calc-browser-pos (fn [x-op x y mult-offset]
                              (if gbrowse
                                (->> (reduce - (sort-by - [x y]))
                                        (double)
                                        (* mult-offset)
                                        (int)
                                       (x-op x))
                                y))
browser-start (calc-browser-pos - padded-start padded-stop 0.2)
browser-end (calc-browser-pos + padded-stop padded-start 0.5)
```

src/datomic_rest_api/rest/fields/gene.clj
```diff
@@ -2085,81 +2087,128 @@
...
-                   "REGULATORY_REGIONS"
-                   "PROMOTER_REGIONS"
-                   "HISTONE_BINDING_SITES"
-                   "TRANSCRIPTION_FACTOR_BINDING_REGION"
-                   "TRANSCRIPTION_FACTOR_BINDING_SITE"
-                   "BINDING_SITES_PREDICTED"
-                   "BINDING_SITES_CURATED"
-                   "BINDING_REGIONS"]}
+  {:data (let [segment (get-segment gene)
+               position (if (empty? segment)
```
mgrbyte: consider chaning to:
```clojure
(if (comp not empty?) segment)
  (segment-to-position gene segment true))
```

src/datomic_rest_api/rest/fields/gene.clj
```diff
@@ -2085,81 +2087,128 @@
...
+  (first
+      (sort-by #(- (:start %) (:end %)) segments)))
+
+(defn- get-segment [gene]
+  (let [segments (get-segments gene)
+        segment (longest-segment segments)]
+    segment))
+
+(defn- segment-to-position [gene segment gbrowse]
+  (let [start (if (< (:start segment) (:end segment))
```
mgrbyte: Find the assignment of `start` and `stop` here confusing (not sure why `:start` can sometime be < `:end` in `segment`?
Anyhow, consider using:
```clojure
[[start, stop] (->> segment ((juxt :start :stop)) (sort-by -)) 
```

mgrbyte (commented):
Generally in `if` forms, no need for `then` _and _`else` in cases when either `then` or `else` is stated as the literal `nil`.
i.e:
;; :/
(if some-condition nil some-value)
;; :)
(if (not some-condition)
    some-value)
```

File: src/datomic_rest_api/rest/fields/gene.clj (hunk 4 above)
Comment (adamjohnwright) [reply src/datomic_rest_api/rest/fields/gene.clj]: The reason has to do with their being positive and negative strands.

File: src/datomic_rest_api/rest/fields/gene.clj (hunk 2 above)
Comment (adamjohnwright) [reply src/datomic_rest_api/rest/fields/gene.clj]: Works for me.

**Comment by sibyl229** [comment]
Hi @a8wright you code looks good to me. I'm okay with it being merged.

File: src/datomic_rest_api/rest/fields/gene.clj [hunk 5]
```diff
@@ -2202,10 +2195,8 @@
 
 (defn feature-image [gene]
   {:data (let [segment (get-segment gene)
-               position (if (empty? segment)
-                          nil
-                          (segment-to-position gene segment true))
-               ]
+               position (if  (comp not empty? segment)
```
mgrbyte: brace in the wrong place here, should be:

```clojure
(if ((comp not empty?) segment)
    ....
```

mgrbyte: This is the only blocker on merging, otherwise looks good to me.

src/datomic_rest_api/rest/fields/gene.clj
```diff
@@ -2085,81 +2087,119 @@
...
+(defn- longest-segment [segments]
+  (first
+      (sort-by #(- (:start %) (:end %)) segments)))
+
+(defn- get-segment [gene]
+  (let [segments (get-segments gene)
+        segment (longest-segment segments)]
+    segment))
+
+(defn- segment-to-position [gene segment gbrowse]
```
mgrbyte, line 2158: minor stylistic issue -  formatting is a bit off here (double spacing in places)

src/datomic_rest_api/rest/fields/gene.clj
```diff
@@ -2085,81 +2087,119 @@
...
-                                     {:text (map #(pack-obj "anatomy-term" (:expr-pattern.anatomy-term/anatomy-term %))
-                                                 (:expr-pattern/anatomy-term expr))
-                                      :evidence {:by (pack-obj "expr-pattern" expr)}})))
-                               (seq))
-               :bound_by (->> (:feature/bound-by-product-of feature)
-                              (map #(pack-obj "gene" (:feature.bound-by-product-of/gene %)))
-                              (seq))
-               :tf  (pack-obj "transcription-factor" (:feature/transcription-factor feature))))))
-       (seq))
+  (let [db (d/entity-db gene)
```
mgrbyte, line 2095: another minor one- perhaps use the required name everywhere when using datomic.api (`d/entity` vs  `entity`, `d/q` vs `q`) (and fixup require not use the :refers). Not technically incorrect, just nice(r) to stick to one way or the other (?)

src/datomic_rest_api/db/sequencesql.clj
```diff
@@ -0,0 +1,5 @@
+(ns datomic-rest-api.db.sequencesql
+  (:require [hugsql.core :as hugsql]))
+
+;(hugsql/def-sqlvec-fns "datomic_rest_api/db/sql/sequence.sql")
```
mgrbyte: nuke commented line?

src/datomic_rest_api/helpers/sequence.clj
```diff
@@ -0,0 +1,9 @@
+(ns datomic-rest-api.helpers.sequence
+  (:require [hugsql.core :as hugsql]
+            [datomic-rest-api.db.sequence :as sequences]
+            [clojure.java.jdbc :as j]
```
mgrbyte: unused required namespaces (jdbc, clojure.string)

Makefile
```diff
@@ -1,6 +1,6 @@
 NAME := wormbase/datomic-to-catalyst
 VERSION ?= $(shell git describe --abbrev=0 --tags)
-WS_VERSION ?= WS256
+WS_VERSION ?= WS257
```
mgrbyte: use `:=` for plain assignment

Makefile
```diff
@@ -18,6 +18,11 @@ need-help := $(filter help,$(MAKECMDGOALS))
 help: ; @echo $(if $(need-help),,\
 	Type \'$(MAKE)$(dash-f) help\' to get help)
 
+.PHONY: get-assembly-json
+get-assembly-json: $(call print-hep,get-assembly-json,\
```
mgrbyte, line 22: Typo here `print-hep`, should be `print-help`

mgrbyte (changes requested):
Tested on the dev server and got json back.

src/datomic_rest_api/db/sequence.clj
```diff
@@ -0,0 +1,48 @@
+(ns datomic-rest-api.db.sequence
+  (:require [datomic-rest-api.db.sequencesql :as sequencesql]
+            [environ.core :refer (env)]
+            [clojure.string :as str]
+            [clojure.data.json :as json]))
+
+(defn database-version []
+  (env :wb-version "WS257"))
```
mgrbyte, line 8: This should be `ws-version` since as per the Makefile using `WS_VERSION`

src/datomic_rest_api/rest/fields/gene.clj
adamjohnwright: Will fix this later - needs a little discussion

mgrbyte (approved):

PR #55, opened 2017-01-31
Feature/re org prep

mgrbyte:
@a8wright @sibyl229  @azurebrd  I've split up my work in to 3 separate PRs, hopefully this will help the merge.
This one is just some basic fixes and lein settings - shouldn't be anything controversial here.


sibyl229:
Hi @mgrbyte Everything looks great to me. A lot more standardized now. @a8wright is prep for another meeting, he says he will look at your PR when he's done (around 1pm). Just one question, not actually a problem, why does the eastwood setting in project.clj need to be quoted?

mgrbyte:
@sibyl229 I'm not sure, perhaps a new version of the plugin caused me to change it to be a string in project.clj.  In this PR, it actually causes an error, so I'll update it (and fix linting errors too).

I'll look into  it when this merged.

sibyl229:
@mgrbyte sounds good!

mgrbyte:
@sibyl229 The answer to the question about quoting the eastwood options stems from [this][1] PR in the `environ` dependency - specifically the `lein-environ` plugin and it's interaction with [squiggly-clojure][2],
which I use with emacs for interactive debugging and code-linting. 

I've fixed this my global [lein configuration][3].
I think rather than repeating this in each clojure project, we should keep editor-specific lein
configuration in each users'  individual `~/.lein/profiles.clj`. 

Unfortunately, this makes development a bit more complicated due to the merging of profiles - but I think editor configuration should up to each developer rather than putting extra dependencies and complexity into the project(s).

I'll open an issue to remove any editor (squiiggly) based config from projects (if there is any).

[1]: https://github.com/weavejester/environ/pull/45
[2]: https://github.com/clojure-emacs/squiggly-clojure
[3]: https://github.com/mgrbyte/dot-files/blob/master/lein/.lein/profiles.clj#L2

sibyl229:
Thanks @mgrbyte for the explanation! Removing editor specific configs from the repo makes sense to me.

PR #77, opened 2017-02-12
Feature/sequence widget

adamjohnwright:
This request contains the sequence URL.

I have tested it with both WB00000001 and abi-1 genes. 

I will test on monday with more genes in order to make sure it is working and then I will set it up so that staging is pointing at it. 

I will assign someone to pull the request after I get a chance to test it


adamjohnwright:
you can look at he results running on http://dev.wormbase.org:9017/species/c_elegans/gene/WBGene00015146#-g-10

Let me know if the page doesn't load. For some reason I have needed to restart the catalyst server a few times.

PR #79, opened 2017-02-14
finished reagent widget - time for curator testing

adamjohnwright:
This contains only the reagent widget. Appears to be working correctly. I will setup dev website to point to it to have curators look at it. 

I have performed the tests and eastwood. The reagent code is passing!

I will assign this pull request when I have a dev environment running off of this code.

### DIFF
--- a/src/rest_api/classes/gene/widgets/reagents.clj
+++ b/src/rest_api/classes/gene/widgets/reagents.clj
@@ -1,0 +1,239 @@
+(ns rest-api.classes.gene.widgets.reagents
+  (:require
+    [datomic.api :as d]
+    [rest-api.classes.gene.generic :as generic]
+    [rest-api.formatters.object :as obj :refer  [pack-obj]]
+    [pseudoace.utils :as pace-utils]
+    [pseudoace.locatables :as locatables]))
+
+(defn- construct-labs [construct]
+  (seq
+    (map
+      #(pack-obj "laboratory" (:construct.laboratory/laboratory %))
+      (:construct/laboratory construct))))
+
+(defn- transgene-labs [tg]
+  (seq (map #(pack-obj "laboratory" (:transgene.laboratory/laboratory %))
+            (:transgene/laboratory tg))))
+
+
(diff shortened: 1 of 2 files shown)


src/rest_api/classes/gene/widgets/reagents.clj
```diff
@@ -0,0 +1,239 @@
...
+          (map
+            (fn [abid]
+              (let [ab (d/entity db abid)]
+                {:antibody (pack-obj "antibody" ab)
+                 :summary (:antibody.summary/text (:antibody/summary ab))
+                 :laboratory (map (partial pack-obj "laboratory") (:antibody/location ab))})))
+          (seq))
+     :description "antibodies generated against protein products or gene fusions"}))
+
+#_(def ^:private child-rule
```
sibyl229: Hi @mgrbyte, your insight on Clojure syntax would be really helpful here. @a8wright and I don't understand what is this `#_`  here that Thomas wrote, and if this has anything to do with the second `(def child-rule)` a few lines later. Have you seen this type of syntax before? Thanks!

sibyl229: @mgrbyte Sorry, never mind that. @a8wright `#_` is the ignore next form macro. https://clojure.org/reference/reader#macrochars  It roughly means this thing is commented out. So you probably need to remove it.

src/rest_api/classes/gene/widgets/reagents.clj
```diff
@@ -0,0 +1,239 @@
...
+     ;;
+     ;; Big assembly-navigation query should probably be factored out somewhere
+     ;; once we're a bit more solid about how this stuff should work.
+     ;;
+     (if parent
+       (->> (d/q '[:find [?p ...]
+                 :in $ % ?seq ?min ?max
+                 :where [?method :method/id "Orfeome"]
+                 (or-join [?seq ?min ?max ?method ?p]
+     (and
```
sibyl229: The indentation here is a bit misleading here.

PR #85, opened 2017-02-16
Reduce required env vars to just the one for datomic URI.

### LINKED ISSUE #58: WS_VERSION environment variable is not needed
Re-implement the function in db.clj which currently uses `(env :ws-version)` to get it from the datomic-uri instead.

mgrbyte:
* Removed the need for WS_VERSION
 * Renamed env-var `TRACE_DB` to `WB_DB_URI`
   * Datomic URI should be specified the same way in all clj projects *
     now.
 * Updated the ElasticBeanstalk .config file accordingly.

Fixes #58

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
 NAME := wormbase/datomic-to-catalyst
 VERSION ?= $(shell git describe --abbrev=0 --tags)
-WS_VERSION := WS257
 LOWER_WS_VERSION = `echo $(WS_VERSION) | tr A-Z a-z`
 EBX_CONFIG := .ebextensions/.config
 DB_URI ?= $(shell sed -rn 's|value:(.*)|\1|p' \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -1,6 +1,5 @@
 NAME := wormbase/datomic-to-catalyst
 VERSION ?= $(shell git describe --abbrev=0 --tags)
-WS_VERSION := WS257
 LOWER_WS_VERSION = `echo $(WS_VERSION) | tr A-Z a-z`
```
adamjohnwright: I am not sure what this line is for, and there appears to be WS_VERSION in this file a couple more places. Any ideas?

mgrbyte: I seem to have missed a thing or to in the Makefile. will fix up the PR.

mgrbyte:
@a8wright I've tested:
  * make get-assemblies-json 
  * make docker-build
  * make docker-run

PR #86, opened 2017-02-16
Was splitting 'publication date' string, and data is empty for papers…

azurebrd:
… like

WBPaper00000974, causing an error.

### DIFF
--- a/src/rest_api/classes/person/widgets/publications.clj
+++ b/src/rest_api/classes/person/widgets/publications.clj
@@ -17,7 +17,9 @@
              (map (fn [oid]
                     (let [paper (d/entity db oid)]
                       (pace-utils/vmap
-                        :year (first (str/split (:paper/publication-date paper) #"-"))
+                        :year (or (if (:paper/publication-date paper)
+                                    (first (str/split (:paper/publication-date paper) #"-")))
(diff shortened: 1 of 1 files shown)


azurebrd:
http://www.wormbase.org/resources/person/WBPerson320#053--10
Has some papers with unknown year showing as if they were grouped with the oldest year, since there's a blank as the key for the year, it doesn't make a year header in the display.  compare with 
http://juancarlos.wormbase.org/resources/person/WBPerson320#053--10

azurebrd:
Emailed Kimberly (Paper curator) for feedback.

vanaukenk:
Thanks @azurebrd .
For now, let's give these papers the heading 'Year Not Available'.
In the meantime, I think I just need to get the list of paper objects with no year and see how many we can populate.  For WBPaper00000974, for example, I was able to add the correct year.

azurebrd:
Thanks @vanaukenk 

Updated and will make a pull request.

If you want to find the papers in postgres, this query should give you them, but it's a slow query to run.
SELECT * FROM pap_status WHERE pap_status = 'valid' AND joinkey NOT IN (SELECT joinkey FROM pap_year);

PR #99, opened 2017-03-12
Feature/external links

adamjohnwright:
This is a result of my work on the variation page. I took some of the ideas there and put them into this code.

Throughout the process I learnt that 17 out of the 35 classes have external links widgets. With this pull request we will be handling all 17 external_links widgets!

## Lessons learnt

There are two types of database links off of entities in Datomic:

e.g. homology group
(schema
  homology-group.database
  (fields
   [database :ref]
   [database-field :ref]
   [text :string :indexed]))

e.g. clone
(schema
  clone.database
  (fields [accession :string :indexed] [database :ref] [field :ref]))

This pull request handles both of these situations. 

I also believe we are handling all possible scenarios where there could be nil values.

Also @mgrbyte and I discussed handling the space in the class names:

The URLs now are routed so that there is an underscore where there would be a space. We convert the underscore to a dash when working with the entity term. At this point only the entity term needs to be specified in the route.

## Testing 

I have queried and compared several URLs per class

## List of supported classes
- Analysis
- Anatomy Term
- CDS
- Clone
- Disease
- Gene (was previously supported but now is supported with reorganized code. 
- Homology Group
- Interaction
- Molecule
- Motif
- Paper
- Protein
- RNAi
- Sequence
- Structure Data
- Transcript
- Variation
 
I created the exteranl


adamjohnwright:
@mgrbyte I know that I should fix this but I tried and was not able to figure it out. Basically one of the tests is not passing. This would be because xrefs was no longer available as a function in the external_links widget on the gene page. I tried adding it back in but that didn't seem to work. I tried changing the test to point at another function and that didn't work. Which makes sense because when I curl the field URI, I do not get a response. I didn't think I changed anything that would break how the fields work. Any ideas why the fields don't seem to be working?

mgrbyte:
@a8wright The description above seems to have been truncated (ends with "I created the exteranl")

I've run the tests (via `lein code-qa`).
There are two main issues:

  1. `NullPointerException` in `test-defroutes`.
       From the error traceback, I was able to determine the cause: 
       now that we're using `str/replace` on the `entity-class` value in a `RouteSpec` mapping, throws a NullPointerException since the test didn't define it previously. Adding a dummy "entity-class" key and value into the route-spec mapping will get rid of the nullpointerexception.

 2. Faulty logic in main implementation of generic/xrefs
     eastwood linting hints at this, when it reports: `suspicious-expression: partial called with 1 args.  (partial f) always returns f.  Perhaps there are misplaced parentheses?`  
    Will comments inline in code review.

src/rest_api/classes/generic.clj
```diff
@@ -0,0 +1,52 @@
...
+          (let  [species-name-parts (str/split species #" ")
+                          g  (str/lower-case  (ffirst species-name-parts))
+                                   species  (second species-name-parts)]
+                (str/join "_"  [g species])))
+
+(defn xrefs [object]
+  (let [data
+        (if-let [k (first (filter #(= (name %) "id") (keys object)))]
+          (let [role (namespace k)
+                ckw (partial (str/join "." [role "database"]))]
```
mgrbyte: This is what `eastwood` is hinting is wrong, it's an expression returns a function, where it's intended usage appears to be as a string (used as first argument to `keyword`), thus I believe should just be `ckw (str role ".database")`.

src/rest_api/routing.clj
```diff
@@ -66,18 +66,19 @@
                        (apply merge (cons fields (->> this :widget vals)))
                        fields)
           route-data (assoc this :field field-defs)
-          entity-class (:entity-class route-data)]
+          entity-class (:entity-class route-data)
+          entity-segment (str/replace entity-class #"-" "_")]
```
mgrbyte, line 67: This is fine - just to note that this is why `rest-api.test-routing/test-defroutes` now needs to supply `entity-class` with a concrete string value in tests.

src/rest_api/routing.clj
```diff
@@ -43,7 +43,7 @@
           entity-class (-> (:context request)
                            (str/split #"/")
                            (last))
-          attr (keyword entity-class "id")
+          attr (keyword (str/replace entity-class #"_" "-") "id")
```
mgrbyte: I think it'd be preferable to pass in a `entity-class` to the `make-request-handler` function as a new parameter; either as is, or just pass the `route-data` mapping, and get `entity-class` from that `(:entity-class route-data)`).

test/rest_api/test_routing.clj
```diff
@@ -31,15 +31,15 @@
       (is (= (:body response) expected))))
   (testing
     "Main request handler produces correct data structure (fields)"
-    (let [req {:uri "/rest/field/gene/WBGene00000001/xrefs"
+    (let [req {:uri "/rest/field/gene/WBGene00000001/rearrangements"
```
mgrbyte: This file should be reverted to use `/xrefs`, with the other changes suggested here, it'll pass again.

mgrbyte (changes requested):
After applying the changes I suggest below, all but one tests pass.

The failing test is `test-alleles-and-polymorphisms` - not that I'm using WS258 to test this, so it could be a data discrepancy:
```
FAIL in (test-alleles-and-polymorphisms) (gene_test.clj:83)
molecular change effects outside of CDS
expected: (some (fn* [p1__15773#] (= "Promoter" p1__15773#)) (:effects allele-other3))
  actual: (not (some #object[rest_api.classes.gene_test$fn__15774$fn__15851 0x67a3713 "rest_api.classes.gene_test$fn__15774$fn__15851@67a3713"] nil))
```

adamjohnwright:
After pulling in @mgrbyte 's changes I have tested with a few urls and it appears to be working. And the tests show 0 Exceptions.

sibyl229:
@a8wright looks good to me! I hope the test case for allele is sorted out for you too. I tried it on develop branch with WS258, it seems working for me.

mgrbyte (approved):

PR #109, opened 2017-05-11
Integration GraphQL routes

sibyl229:
Not for merging yet. But it's a working copy the code for anyone interested to try out GraphQL with WormBase data, through a thin layer of abstraction on top of Datomic database.

I wonder what people think of using routes (ring handlers) defined in another package (wb-graphql) and the idea of GraphQL service tagging along like this. The objective here is to avoid the devOp overhead of having another independent web service.


mgrbyte:
@sibyl229 I think generally having other services mixed together to avoid the overhead of dev-ops is undesirable.  I do understand that the WB may be constrained in terms of resource, so might be unavoidable.

I'm concerned about this approach; mainly, but not limited to performance.
Memory usage profile of the application could be adversely affected.

I think having independent web services allows for independent replacement and management from ops perspective, and less moving parts in a single application; less things to consider, which I think are desirable properties.

Could you expand on what dev-ops overhead you're attempt to mitigate? i.e resourcing issues (e.g AWS) vs procedural (tooling)?

project.clj
```diff
@@ -71,4 +72,5 @@
           :ring {:nrepl {:start? true}}}]
       :test
       {:resource-paths ["test/resources"]}}
-  :repl-options {:init (set! *print-length* 10)})
+  :repl-options {:init (set! *print-length* 10)}
+  :repositories [["tmp" "file:///tmp/clojure_jars"]])
```
mgrbyte, line 76: Can you explain what this repository in /tmp is used for? (Don't think /tmp is the best place - won't survive reboot!)

mgrbyte (commented):
Nothing wrong with code itself.

sibyl229:
Closing. This needs to be better thought out.

PR #112, opened 2017-05-19
feature/gene interactions widget

mgrbyte:
This implements the gene interactions widget.

The widget takes a very long time to render, depending on the gene in focus.
The results could be [cached][1] in memory, using an LRU strategy by wrapping the functions that produce the views.
Citations with between 1 and 6 authors have now been standardised to the format:
   `<author1>, <author2> & <author3>, <date>`
    
[1]: https://github.com/clojure/core.memoize


src/rest_api/classes/gene/widgets/interactions.clj
```diff
@@ -91,16 +54,67 @@
...
+(defn- interaction-type-name [interaction]
+  (let [itype (first (:interaction/type interaction))
+        type-name (humanize-name itype)]
+    (cond
+      ;; Hack to produce the "type-name" when real type-name
+      ;; is regulatory.
+      ;; TODO: Perhaps the proposed module system should be able to address
+      ;;       this and produce a cleaner solution.
+      (str/includes? (str/lower-case type-name) "regulat")
+      (if-let [reg-res (regulatory-result interaction)]
```
adamjohnwright: I have started using when-let instead of if-let when possible. The difference is that when-let does not have an "else" part. This makes it a bit more readable because you know that there is no "else" to the condition.

src/rest_api/formatters/object.clj
```diff
@@ -160,18 +161,17 @@
   ;; computing the display name.
   (let [db (d/entity-db int)]
     (if-let [il (seq (d/q q-interactor db (:db/id int)))]
```
adamjohnwright, line 163: Same here - consider "when-let"

adamjohnwright (changes requested):

src/rest_api/formatters/object.clj
mgrbyte: This one has an else clause, so haven't changed it.

sibyl229 (approved):
Code is good. Such a complex widget though. Feels that something else is off when complex logic is needed to produce widget JSON.

azurebrd (approved):

src/rest_api/formatters/object.clj
mgrbyte: @a8wright  Changes done in 0f02db0dedf72b26f21288954907c0ab63eeb639.

adamjohnwright (approved):

PR #117, opened 2017-06-13
Bugfix/interaction details performance fix 1

mgrbyte:
Addresses performance issues for genes that have few interactions.

The queries the code were making to datomic for the "nearby" interactions were not specified correctly, causing the equivalent of a "full-table scan" in SQL - i.e was consuming all interaction entities  and filtering rather than putting the most selective statement first.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -55,7 +55,8 @@
            ["do"
             ["eastwood"]
             "test"]}
-          :dependencies [[ring/ring-devel "1.5.1"]]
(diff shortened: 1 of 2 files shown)


sibyl229:
@mgrbyte this makes a lot of sense! Great catch. Since it's kinda urgent (for WS259), I will merge this into develop so we can see it in action (before @a8wright has a chance to look at it. Sorry @a8wright )

PR #118, opened 2017-06-20
Swagger doc for id param

sibyl229:
Hi @mgrbyte , I added a description for the `:id` url parameter to the Swagger doc. Are you okay with the change?

screen shot 2017-06-20 at 4 35 42 pm

Cheers!

### DIFF
--- a/src/rest_api/routing.clj
+++ b/src/rest_api/routing.clj
@@ -6,7 +6,6 @@
    [rest-api.db.main :refer [datomic-conn]]
    [ring.util.request :as req]
    [ring.util.http-response :as res]
-   [ring.swagger.schema :as swagger-schema]
    [schema.core :as schema]))
 
 (defn- entity-not-found [schema-name id]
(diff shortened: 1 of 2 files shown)


mgrbyte:
@sibyl229 Looks great :+1:

src/rest_api/routing.clj
```diff
@@ -55,6 +55,10 @@
     [route-spec]
     [route-spec opts]))
 
+(defn swagger-id-field [entity-type]
```
mgrbyte: perhaps use `entity-name` or maybe `entity-alias` here instead of `entity-type` (the "catalyst" name of the entity in the URL (e.g "disease" instead of "do-term" et al).

mgrbyte (approved):
Looks good.

src/rest_api/routing.clj
sibyl229: Ah Thanks @mgrbyte ! Can I name it `entity-type-alias`? Trying to convey that it's the alias of a class (gene) rather than an instance (WBGene000...)

mgrbyte: I'd be fine with `string` or `s` too.

mgrbyte (approved):

PR #123, opened 2017-07-19
Feature/references widget

adamjohnwright:
This includes 22 references widgets for the following pages:

- analysis
- anatomy_term
- antibody
- cds
- clone
- construct
- expression_cluster
- expr_pattern
- expr_profile
- gene
- gene_cluster (This is a brand new widget)
- life_stage
- molecule
- operon
- rearrangement
- rnai
- sequence
- strain
- transcript
- transgene
- variation
- wbprocess

I am not sure how many more are brand new widgets but I think there must be a few more than the one marked (gene_cluster).

Let me know what you think. I walk over to talk over a plan with you.


src/rest_api/classes/anatomy_term.clj
```diff
@@ -1,13 +1,16 @@
...
-    ;[rest-api.classes.anatomy_term.widgets.overview :as overview]
-    [rest-api.classes.anatomy_term.widgets.ontology-browser :as ontology-browser]
+    ;[rest-api.classes.anatomy-term.widgets.overview :as overview]
+    [rest-api.classes.anatomy-term.widgets.references :as references]
+    [rest-api.classes.anatomy-term.widgets.ontology-browser :as ontology-browser]
     [rest-api.routing :as routing]))
 
 (routing/defroutes
-  {:entity-ns "anatomy-term"
+  {:entity-ns "anatomy-function"
```
mgrbyte: Just curious; do you know why the api for anatomy term uses the `anatomy-function` as the "entry entity" ?

src/rest_api/classes/generic_fields.clj
```diff
@@ -178,3 +178,52 @@
...
+                                                    (for [h hs
+                                                          :let [author (cond
+                                                                         (contains? h :affiliation/person)
+                                                                         (first (:affiliation/person h))
+
+                                                                         (contains? h :paper.author/author)
+                                                                         (:paper.author/author h))]]
+                                                      {(:ordered/index h)
+                                                       (pack-obj author)})))))
+                                    :ptype (if (nil? paper) nil (:paper.type  pt))
```
mgrbyte: Two things with the `ptype` key: 
1) I think the key you want is `:paper.type/type` (:paper.type is the component schema)
2) Can more simply write `(when paper (:paper.type/type pt))` or `(some-> pt :paper.type/type)`

src/rest_api/classes/generic_fields.clj
```diff
@@ -178,3 +178,52 @@
...
+                                    :abstract (if (nil? abstract) nil [(:longtext/text (first abstract))])
```
mgrbyte: Same comment as above, rather than having to code both clauses of `if`, just have:
```clojure
(when abstract
   [(:longtext/text (fist abstract))])
```

mgrbyte (changes requested):
All looks ok bar places where you have the pattern :
`(if (nil? thing) nil (do-something thing)`

Either:
`(some-> thing (do-something thing)`
or
`(when (not-empty thing) (do-somthing thing)`

would be preferable. What do you think?

src/rest_api/classes/anatomy_term.clj
adamjohnwright: I do not. I think it would be interesting to find out about the schema to find out why it is the way it is. I wish I had deeper insight

src/rest_api/classes/generic_fields.clj
adamjohnwright: I agree. The reason the code is this way is because I wrote it a while ago. I didn't change the core functionality from it. That being said. Sibyl will be going over what I am data I am providing and will likely be asking for changes. The data for this widget is actually not coming directly from ACeDB. It is comming from the search index. So there were probably be improvements that Sibly will be wanting. Thanksf for pointing this out though. I will fix this the next time I touch the code.

src/rest_api/classes/generic_fields.clj
adamjohnwright: same as above. I will let you know when the changes have been made.

adamjohnwright:
Hi Matt, 

I have gone through all of your requests. I think that everything is resolved. I will be asking Todd on this weeks call about the anatomy vs anatomy-function entity type for the anatomy page.

src/rest_api/classes/generic_fields.clj
```diff
@@ -12,22 +13,22 @@
   (let [id-kw (first (filter #(= (name %) "id") (keys object)))
         role (namespace id-kw)]
     {:data (when-let [ghs ((keyword role "gene") object)]
-             (for [gh ghs :let  [gene ((keyword (str role ".gene") "gene") gh)]]
+             (for [gh ghs :let [gene ((keyword (str role ".gene") "gene") gh)]]
```
mgrbyte: Perhaps this shouldn't block merging, but minor style request to keep `:let` declarations in a `for` for it's own line, e.g:
```clojure
(for [x y
     :let [foo bar]]
   (do-stuff ...))
```

mgrbyte: Also, in the `for`, `(keyword (str role ".gene")` could be pulled up to a var in the outer let to avoid recomputing it each time.

mgrbyte (changes requested):
Added a couple more items I missed on first-pass.
Mainly: move-up keyword creation to outer-lets where applicable.

src/rest_api/classes/anatomy_term.clj
sibyl229: @a8wright I don't think this is correct. The Anatomy page on wb site definitely corresponds to anatomy term ace object. You can see it by going to an anatomy page and see the ace tree view: http://www.wormbase.org/species/all/anatomy_term/nerve%20ring%20neuron#014--10

sibyl229: anatomy-function is something else

sibyl229 (changes requested):

mgrbyte (approved):
As discussed on slack; using combinations of `->>` or `some->>` with `map` preferable to using `for` - but don't think this should block merging.
Approving this, as the question per the distinction between `anatomy-term` and `anatomy-function` should be resolved by @sibyl229 's  review,

sibyl229 (approved):

PR #126, opened 2017-07-27
adding location widgets

adamjohnwright:
This pull request contains location widgets for:
- cds
- clone
- feature
- operon
- pcr_product (commented out)
- protein
- pseudogene
- sequence
- transcript
- transposon
- variation

The pcr_product widget is not being released because I noticed and error when running the following command:

```bash
curl -g -H content-type:application/json http://localhost:8130/rest/widget/pcr_oligo/cenix:10-a2/location | jq '.
```
Basically my code requires the species field but for this entity it was not available. 

I realise that for protein there still requires some work. I will be doing one more pass through tomorrow

### DIFF
--- a/src/rest_api/classes/expression_cluster/widgets/overview.clj
+++ b/src/rest_api/classes/expression_cluster/widgets/overview.clj
@@ -19,8 +19,7 @@
 	      (let [id-kw (first (filter #(= (name %) "id") (keys object)))]
 		{(let [i (id-kw object)]
 		   (if-let [result (re-find #":(\d+)_min" i)]
-		     (do (println (second result))
-			 (pad-zero (second result) 10)) i))
+			 (pad-zero (second result) 10)) i)
 		 (pack-obj object)}))))))
 
 (defn attribute-of [ec]
--- a/src/rest_api/classes/cds.clj
+++ b/src/rest_api/classes/cds.clj
@@ -2,6 +2,7 @@
   (:require
     [rest-api.classes.gene.widgets.external-links :as external-links]
     ;[rest-api.classes.cds.widgets.overview :as overview]
+    [rest-api.classes.cds.widgets.location :as location]
     [rest-api.classes.cds.widgets.references :as references]
     [rest-api.routing :as routing]))
 
@@ -9,5 +10,6 @@
   {:entity-ns "cds"
    :widget
    {;:overview overview/widget not complete -  has some calculations that would need to be verified
+    :location location/widget
     :external_links external-links/widget
     :references references/widget}})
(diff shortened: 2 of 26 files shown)


adamjohnwright:
Just went through this today. The main work that was required was to get it working correctly for protein. I think I am now replicating all the logic in the Perl code. 

The plan for testing is to get this in to staging and then get someone like Scott Cain to look it over.

So far I have gone through several endpoints for each entity type and everything looks good.

sibyl229 (approved):
I think it's okay for merging. But I will add a bit of my thoughts on the implementation of the genetic-position function. 

The function becomes more complex and opaque, in attempt to handle different input types (such as variations and proteins). If we encounter similar situation in the future, I think maybe we could consider providing multiple implementation of the same function based on input type.

Another more subjective opinion is that dynamically constructed keywords makes code harder to read. For example
```
+                     (let [kw-imp (keyword entity-role str-imp)
 +                           str-imp-component (str entity-role "." str-imp)
 +                           kw-map (keyword entity-role "map")
 +                           kw-map-map (keyword (str entity-role ".map") "map")
 +                           kw-imp-map (keyword str-imp-component "map")
 +                           kw-imp-position (keyword
 +                                             str-imp-component
 +                                             (if (= role "sequence") "float" "position"))]
```
vs 
```
:variation/interpolated-map-position
:varitation.interpolated-map-position/map
```

src/rest_api/classes/expression_cluster/widgets/overview.clj
```diff
@@ -19,8 +19,7 @@
 	      (let [id-kw (first (filter #(= (name %) "id") (keys object)))]
 		{(let [i (id-kw object)]
 		   (if-let [result (re-find #":(\d+)_min" i)]
-		     (do (println (second result))
-			 (pad-zero (second result) 10)) i))
+			 (pad-zero (second result) 10)) i)
```
mgrbyte, line 22: this needs dedenting

mgrbyte (approved):
Generally, It looks good to merge.

Agree with comments @sibyl229 made about the dynamically constructed keywords (albeit I'm probably guilty of introducing that pattern!). 
The above (dynamic kw construction) could be discussed further, I think it's combination of the nature of the dynamic schema conversion + clojure "foo" (e.g perhaps we could look further into clojure multimethods/protocols vs dynamically handling things based on a "role" string) for this type of generic dispatch.

adamjohnwright:
BTW. I agree that over time we should be coming up with better patterns for doing these things. I would be more than willing to make better code. I am not sure how / when we should do this but we should keep this in the back of our minds.

PR #127, opened 2017-08-14
Migrate ACeDB dependency of BLAST result to d2c

sibyl229:
Objective: remove ACeDB dependency from formatting of BLAST result

IDs of BLAST hits need to be converted to the `pack-obj`format of their corresponding gene and protein for display. Catalyst app currently relies on ACeDB for this conversion, resulting in frequent crash of BLAST service. https://github.com/WormBase/website/blob/master/lib/WormBase/API/Service/blast_blat.pm#L709-L727

Solution:  create an additional endpoint for ID conversion in the d2c repo

Technical details:
- additional route `/rest/convert/:id` is added to `ns rest-api.classes.blast-hit`
- multimethod is used to handle IDs of different types (like :protein/id vs :transcript/id).

Thanks!

### DIFF
--- a/src/rest_api/classes/blast_hit.clj
+++ b/src/rest_api/classes/blast_hit.clj
@@ -41,15 +41,6 @@
      :protein (pack-obj "protein" protein :label "[Protein Summary]")
      :sequence (pack-obj cds)}))
 
-;; est
-(defmethod convert-entity :sequence/id [sequence]
-  (let [gene (some->> sequence
-                      (:locatable/_parent)
-                      (filter :gene/id)
-                      (first))]
-    {:corresponding_gene (pack-obj "gene" gene :label "[Corr. Gene]")
-     :sequence (pack-obj sequence)}))
-
 (defmethod convert-entity :default [entity]
   {:sequence (pack-obj entity)})
 
--- a/src/rest_api/classes/blast_hit.clj
+++ b/src/rest_api/classes/blast_hit.clj
@@ -13,6 +13,7 @@
   (fn [entity]
     (first (filter (partial contains? entity) id-types))))
 
+;; wormpep
 (defmethod convert-entity :transcript/id [entity]
   (let [transcript entity
         protein (some->> transcript
@@ -28,6 +29,7 @@
      :protein (pack-obj "protein" protein :label "[Protein Summary]")
      :sequence (pack-obj transcript)}))
 
+;; wormpep
 (defmethod convert-entity :cds/id [entity]
   (let [cds entity
         protein (some->> cds
@@ -41,6 +43,7 @@
      :protein (pack-obj "protein" protein :label "[Protein Summary]")
      :sequence (pack-obj cds)}))
 
+;; est
 (defmethod convert-entity :sequence/id [entity]
   (let [sequence entity
         gene (some->> sequence
--- a/src/rest_api/classes/blast_hit.clj
+++ b/src/rest_api/classes/blast_hit.clj
@@ -7,10 +7,12 @@
    [ring.util.http-response :as response]
    [schema.core :as schema]))
 
+(def id-types [:transcript/id :cds/id :sequence/id :protein/id])
+
 (defmulti convert-id
   (fn [id]
     (let [db (d/db datomic-conn)]
-      (->> [:transcript/id :cds/id :sequence/id :protein/id]
+      (->> id-types
            (filter #(d/entity db [% id]))
            (first)))))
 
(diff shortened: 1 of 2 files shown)


mgrbyte (commented):
Overall I really like the approach :+1: .

I would make one minor adjustment:

Change the signature of `convert-id` to accept `entity` instead of `id` to avoid repetition of the get-db-from-connection-then-lookup-entity in each `defmethod`.
If you think it's a good to adopt this idea, the `defmulti` discriminator fn would change to:
```
(fn [entity]
  (first (filter (partial contains? entity) id-types))
``` 
and the call site:
```
(response/ok (convert-id (some #(d/entity db [% id]) id-types))
```
The bulk of the rest of the code would remain as is,  changes would be:
  * remove lookup of the entity in each `defmethod`
  * The `:default` implementation would simplify down to `(defmethod convert-id :default [sequence]
    {:sequence (pack-obj sequence)}`. 

Only "problem" with this idea is the `nil` case, which I have a question about.
If an id does not resolve to an entity matching one of the `id-types`, currently the response will be 200 OK with a JSON body `{"sequence" {"label" "whatever-the-user-entered-as-an-id-in-the-URL"}}`;
should that not be a 404? or is this for backwards-compatibility behaviour with the old endpoint?

adamjohnwright (approved):

sibyl229:
Thanks a lot @mgrbyte ! All suggestions are adopted. Passing entity in as param is a great idea. And for the `nil` situation, it's now responding with 404. (I initially had it there for backward compatibility, but it's no longer necessary now).

Thanks @a8wright

azurebrd (approved):
Thanks, didn't know we could make a db connection through [datomic-conn] instead of requiring an entity.

src/rest_api/classes/blast_hit.clj
```diff
@@ -0,0 +1,66 @@
...
+
+(def id-types [:transcript/id :cds/id :sequence/id :protein/id])
+
+(defmulti convert-entity
+  (fn [entity]
+    (first (filter (partial contains? entity) id-types))))
+
+;; wormpep
+(defmethod convert-entity :transcript/id [entity]
+  (let [transcript entity
```
mgrbyte: redundant alias; rename the `entity` parameter to `transcript` or just use `entity`

src/rest_api/classes/blast_hit.clj
```diff
@@ -0,0 +1,66 @@
...
+                      (:gene.corresponding-transcript/_transcript)
+                      (first)
+                      (:gene/_corresponding-transcript))]
+    {:gene (pack-obj "gene" gene :label "[Gene Summary]")
+     :protein (pack-obj "protein" protein :label "[Protein Summary]")
+     :sequence (pack-obj transcript)}))
+
+;; wormpep
+(defmethod convert-entity :cds/id [entity]
+  (let [cds entity
```
mgrbyte: Same as above (parameter aliasing)

src/rest_api/classes/blast_hit.clj
```diff
@@ -0,0 +1,66 @@
...
+                      (:gene.corresponding-cds/_cds)
+                      (first)
+                      (:gene/_corresponding-cds))]
+    {:gene (pack-obj "gene" gene :label "[Gene Summary]")
+     :protein (pack-obj "protein" protein :label "[Protein Summary]")
+     :sequence (pack-obj cds)}))
+
+;; est
+(defmethod convert-entity :sequence/id [entity]
+  (let [sequence entity
```
mgrbyte: drop param alias (same as above)

mgrbyte (changes requested):
Glad you liked & implement the idea @sibyl229 :+1: 
Last few clean-up things!

sibyl229:
Thanks @mgrbyte ! Done. Cheers :)

@azurebrd thanks. And yup, in fact that bit of logic is also at the heart of the route handler / controller : https://github.com/WormBase/datomic-to-catalyst/blob/develop/src/rest_api/routing.clj#L41-L45

mgrbyte:
@sibyl229 good spot that the connection is made when using `rest-api.routing/defroutes`.
This new blast-hit endpoint wants to be non entity-specific, and has custom compojure-api swagger metadata, so don't see the harm in replicating that little bit of logic here.
As a "better" way, we could conider adding ring middleware that annotates the request with the connection and db.
Similar to the not-found middleware in `rest-api.main`:

```clojure
(defn- wrap-datomic
  "Annotated request with datomic connection and current db."
  [handler]
  (fn [request]
    (handler (assoc request :conn datomic-conn (d/db datomic-conn)))))
```

mgrbyte (approved):
Great work! :+1:

sibyl229:
Thank you @mgrbyte ! The middleware approach is neat. Definitely something we can try in the future. Especially when we have history in the database, can use the middleware to select right database.

I'm gonna merge for now. So things can go into WS260 :)

PR #128, opened 2017-08-19
Lineage widget. Tables for supervised, supervised_by, worked_with.

azurebrd:
ancestors_data json for direct and full person lineage cytoscape viewer.


azurebrd:
Please let me know if there's indenting or other formatting issues.  Also I'm not sure why the overview widget is commented out in develop, causing the merge conflict.  I thought it went live with the last release.  Thanks all.

azurebrd:
Conflict created here
https://github.com/WormBase/datomic-to-catalyst/commit/eeab1ee9f3770191938f580ef3dc280325922f0c
but I don't know why it was changed.

adamjohnwright:
Hi @azurebrd I was likely the one that commented out the overview widget. I think I made the change when releasing the WS250 database. The reason was that I am pretty sure I saw address' that were still not in order. If with WS260 the address' are in order feel free to uncomment. Sorry about the process not being clear. I was just being cautious when releasing them. Hopefully the spreadsheet that I made that is connected to the general channel on slack will help clear up these issues in the future.

adamjohnwright (approved):

azurebrd:
Thanks @a8wright, I'll take a look at the address and old address !

azurebrd:
Hey @a8wright , I haven't had a chance to update the address info yet, I'm just trying to resolve the merge conflict.  When I put the ; in front of the lines to match the person.clj in the develop branch, the overview widget gives a 500 error instead of defaulting to the ACeDB data.  Is that how it's supposed to be because my private instance doesn't go to ACeDB, but the develop / web versions of wormbase would go to ACeDB ?  I'm not sure if I should put the ; to match and commit that despite the 500 error.  If it's possible to merge this branch into develop (assuming no one else suggests changes) and at that point add the ; like you did back then, that'd be great.  If not, please advise on how to solve this merge conflict.  Thanks !

adamjohnwright:
Hi @azurebrd I can see how that could be confusing. The reason for this is because @sibyl229 and I changed the way the catalyst app knows which URI's to send to Datomic v.s. AceDB. Before it was trying Datomic for all the URI's if it didn't get a 200 response from Datomic it would try AceDB. This wasn't really a problem but wasn't quite right. Partly it wasn't right because we didn't want it to fall back to AceDB when we thought it was actually using Datomic. But also it was another http call for every URI that was not provided by Datomic. So what we are now doing is getting a JSON file, periodically, that says what all the URI's are that are provided by Datomic. I am not sure how often it does this but I think it is every 30 seconds or something. So when you comment out your URI you will need to wait a short while for the JSON file to be updated. In short, until it updates it will be giving a 500 error... this is expected behaviour.

azurebrd:
Thanks @a8wright !  I've updated to match the version on develop with commented out overview lines, but also with new lineage lines, and I see it loading from ACeDB.  I've written you on slack with github questions, to avoid a long exchange here that other reviewers might not want to read, but I can paste it as a comment in this ticket, if that's more appropriate.

azurebrd:
Thanks @a8wright for the skype call resolution.  Have a great vacation !

mgrbyte (changes requested):
Generally looks good, but if you could make the following changes it would help keep conventions:

* Use hyphen-style for naming instead of camel case.
   e.g: this-and-that instead of thisAndThat
* separate long names in datalog queries, or use shorter names (I prefer the later, but either would be fine)
   e.g ?personsupervisedby -> ?person-supervised-by  or ?psb
* Use the aliases brought in by require
  e.g clojure.string/lower-case -> str/lower-case

* Use `not-empty` instead of `(if x `nil` y)`, will add examples below inline

* consider using threading macro instead of nested fn calls where it gets unwieldy to read (will add inline example below)

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -0,0 +1,349 @@
...
+              :let [rfrom (keyword
+                            (str
+                              (clojure.string/replace 
+                                (clojure.string/replace 
+                                  (str role) #"/" ".") #":" "") "/from"))]
+              :let [rto (keyword
+                          (str
+                            (clojure.string/replace 
+                              (clojure.string/replace 
+                                (str role) #"/" ".") #":" "") "/to"))]
```
mgrbyte: example where this could threading macro instead:
```clojure
:let [r (-> role (str/replace #"/" ".") (str/replace #":" "")
      rto (str r "/to")]
```

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -0,0 +1,349 @@
...
+              :let [from (if (contains? (first (jsondata role)) rfrom)
+                           (date/format-date5 ((first (jsondata role)) rfrom))
+                           nil )]
+              :let [to (if (contains? (first (jsondata role)) rto) 
+                         (date/format-date5 ((first (jsondata role)) rto)) 
+                         nil )]
+              :let [duration (str from " - " to)]] duration)]
```
mgrbyte: These three lets could/should just be one:
```clojure
:let [from ..
      to ...
      duration ...]
```

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -0,0 +1,349 @@
...
+(defn supervised [person]
+  (let [db (d/entity-db person)
+        data (->> (d/q '[:find ?otherpersonid ?otherpersonname (pull ?personsupervisedbyent [*])
+                         :in $ ?person
+                         :where
+                         [?otherpersonent :person/supervised-by ?personsupervisedbyent]
+                         [?personsupervisedbyent :person.supervised-by/person ?person]
+                         [?otherpersonent :person/id ?otherpersonid]
+                         [?otherpersonent :person/standard-name ?otherpersonname]
+                         ]
```
mgrbyte: minor niggle: it's idiomatic in lisp/clojure to not have trailing braces/brackets on new lines (pull up)

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -0,0 +1,349 @@
...
+    (for [level levels
+          data (if ((keyword level) roleColour) 
+                 (vector
+                   (hash-map
+                     :personid personid
+                     :personname personname
+                     :otherpersonid otherpersonid
+                     :otherpersonname otherpersonname
+                     :level level))
+                 nil )] data)))
```
mgrbyte: Don't need to specify nii explictly, use `if-not` instead

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -0,0 +1,349 @@
...
+                  (if (nil? ((keyword (:otherpersonid value)) scalingHash))
+                    1 
+                    (parse-int ((keyword (:otherpersonid value)) scalingHash))))]] data)))
+
+(defn- filter-this-node [directOrFull queriedData]
+  (let [value (first queriedData)
+        scaling (if (nil? ((keyword (:personid value)) scalingHash))
+                  1 
+                  (parse-int ((keyword (:personid value)) scalingHash)))
+        data (str "id: '" directOrFull (:personid value) "', name: '" (:personname value) "', url: '" (:personid value) "', scaling: '" scaling "', radius: '100', nodeshape: 'rectangle'")]
```
mgrbyte: split long line, e.g:
```clojure
  data (str "id : '"
                direct-or-full
                (:person-id value)
                ...)
```
Perhaps consider using `clojure.core/format` instead of `str` here?

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -0,0 +1,349 @@
...
+        :let [data (str (:level value))]] data))
+
+(defn- filter-existing-roles [existingRolesUnique]
+  (apply merge (for [role existingRolesUnique
+                     :let [data  
+                           (pace-utils/vmap
+                             (keyword role) ((keyword role) roleColour))]] data)))
+
+(defn- filter-edges-supervisees [directOrFull queriedData]
+  (if (empty? queriedData)
```
mgrbyte: could be simplified to:
```clojure
(if queried-data
      (for ...))
```

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -0,0 +1,349 @@
...
+                                   (flatten))]
+    (if (empty? queriedDataThisPerson) nil queriedDataThisPerson)))
+
+(defn ancestors-data [person]
+  (let [db (d/entity-db person)
+        queriedDataThisPerson (query-this-person person)
+        thisPersonId (:personid (first queriedDataThisPerson))
+        thisPersonSupervisors (query-supervisors thisPersonId)
+        thisPersonSupervisees (query-supervisees thisPersonId)
+        edgesDirect
```
mgrbyte: if I understand correctly, the code below is building up a JSON structure?
If so, then better to build the data structure in clojure maps/vectors then pass the result to a json library fn, such as `cheshire.core/generate-string`

azurebrd:
Thanks @mgrbyte I appreciate the separate list of issues and individual examples.  I'll go through them and make changes, possibly asking about them individually if I'm unclear.

Since merging the develop branch to get rid of the conflict yesterday, I'm seeing this issue.  I don't know if it's caused by the merge or that's just a coincidence.  When starting datomic-to-catalyst with

cd ~/datomic-to-catalyst
source ~/aws_creds
export WS_VERSION=WS259
export TRACE_DB="datomic:ddb://us-east-1/WS259/wormbase"
lein ring server-headless 4001

I get these errors that haven't shown up before

Could not find artifact com.datomic:datomic-pro:jar:0.9.5561.50 in central (https://repo1.maven.org/maven2/)
Could not find artifact com.datomic:datomic-pro:jar:0.9.5561.50 in clojars (https://clojars.org/repo/)
This could be due to a typo in :dependencies or network issues.
If you are behind a proxy, try setting the 'http_proxy' environment variable.

Does anyone know how I could try to resolve them ?

mgrbyte:
@azurebrd
datomic-pro version has been bumped to a newer version after you performed the merge from develop to fix merge conflict.

Assuming you have the `feature/person-overview` branch checked out:
```bash
git fetch
git pull
git diff f5a530c..HEAD project.clj
```
Each time we upgrade the version of datomic-pro, you'll need to download and [install](http://docs.datomic.com/integrating-peer-lib.html#local-maven-repository) datomic-pro to update your development environment; we do this fairly regularly (typically between 1->3 releases of dtc have seen an upgrade to a newer version of datomic-pro).

azurebrd:
Awesome, thanks @mgrbyte !  I got the version from /home/ec2-user/datomic-pro/datomic-pro-0.9.5561.50.zip

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: Is this the example referred to above that should be using not-empty ? I'd seen code in the format (if x nil y) and just replicated it.  I've changed it to the syntax in this example, and it seems to run well in my test cases.  Should it be using not-empty in some way instead ?

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: Fixed, thanks.

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: Fixed, thanks.

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: Fixed.

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: I see that it works, and it's easier to follow having the function next to the parameters, but is it better to have that extra variable as opposed to : 
```clojure
:let [rto (keyword (str (-> role (str/replace #"/" ".") (str/replace #":" "")) "/to"))]
```
And in that case, should the (keyword  and (str  be in their own lines ?  I'm not sure what would be best to follow conventions and have good readability.

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: Fixed.  I looked into clojure.com/format, but it didn't seem to make it more readable.

adamjohnwright:
Just a small comment. When using the function "str" make sure the value exists. This can either be done by knowing that the value always exists, through some sort of reasoning, or doing a check in the code. If the function "str" encounters a nil value when it is trying to concatenate d2c will throw a 500 error. From what I have seen in your code most of the places that you use the "str" function the value should logically exist.

azurebrd:
Thanks @a8wright I'll double check that !

After double checking, it looked okay, but I found a new edge case when data is wrong and fixed it.

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: Picked a way of doing the threading macros and changed the code to match.

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: That's true, and that makes sense, I didn't know that was an option when I wrote it.  If it's okay to keep it as it is, then it's working fine, if not I'll think over how to pass the structures around to generate the maps/vectors.  In the future I can write new code with generate-string.

azurebrd:
@mgrbyte I've made a lot of changes, and responded on each of the points.  Some had questions, but basically everything works as is.  Let me know if you'd like further changes, thanks !

mgrbyte:
@a8wright FWIW, your comment about `str` concatenation with `nil` isn't correct, you must have gotten confused with `clojure.string/join` or some other operation: 
```
user> (str "concatenation with " nil " does not cause a null " nil "pointer exception")
"concatenation with  does not cause a null pointer exception"
```

mgrbyte (approved):

mgrbyte:
@azurebrd thanks for making the changes, I like :+1: 
I think it would be good to refactor to use `generate-string`, for performance not just stylistic reasons.
(Tested with several WBPerson ids locally with a running instance, `WBPerson625` which seemed a little slow)

azurebrd:
Cool, thanks @mgrbyte !  I appreciate the feedback, I think the code is much improved from it.

That's interesting, it's a bit slower to run off Datomic than the ACeDB version on the live website.  People like WBPerson625 have lots of nodes and edges, especially in the Full View, and will always take a while to load, but I think it's a pretty reasonable load time.  As I get more used to manipulating data in Clojure on other widgets, I'll come back to this with a clearer way on how to do it.

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -0,0 +1,383 @@
...
+        (for [role roles
+              :let
+              [level (-> role
+                         (str/replace #":role/" "")
+                         (str/replace #"-" " ")
+                         (str/capitalize))]] level)
+        duration
+        (for [role roles
+              :let [rfrom (keyword
+                            (str
```
mgrbyte: I missed this in my review - use of keyword here is a bit off:
should be changed to:
```clojure
(keyword (-> role (str/replace #"/" ".") (str/replace #":" "")) "to")
```

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -0,0 +1,383 @@
...
+                         :where
+                         [?person :person/supervised-by ?person-supervised-by-ent]
+                         [?person-supervised-by-ent :person.supervised-by/person ?other-person-ent]
+                         [?other-person-ent :person/id ?other-person-id]
+                         [?other-person-ent :person/standard-name ?other-person-name]]
+                       db (:db/id person))
+                  (map (fn [oid]
+                         (generate-map oid)))
+                  (seq))]
+    {:data (if data data)
```
mgrbyte: The `if` here is redundant (since `seq` has been applied in the expression which binds `data` above).
i.e `(seq empty-collection)` returns `nil`

mgrbyte (changes requested):
@azurebrd a couple more minor tweaks which I missed them on first pass

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: Ah, that's neat that it then doesn't need the /
Fixed.

src/rest_api/classes/person/widgets/lineage.clj
azurebrd: Cool, that's good to know.  Fixed.

azurebrd:
Thanks @mgrbyte changes pushed.

mgrbyte (approved):
:+1:

adamjohnwright:
Let me know if this should be rolled out into production.                                                                                                                                                                                                                                                                                                                                        Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                From: Juancarlos ChanSent: Wednesday, September 13, 2017 5:24 PMTo: WormBase/datomic-to-catalystReply To: WormBase/datomic-to-catalystCc: Adam; MentionSubject: Re: [WormBase/datomic-to-catalyst] Lineage widget. Tables for supervised, supervised_by, worked_with. (#128)Merged #128.

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/WormBase/datomic-to-catalyst","title":"WormBase/datomic-to-catalyst","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/WormBase/datomic-to-catalyst"}},"updates":{"snippets":[{"icon":"DESCRIPTION","message":"Merged #128."}],"action":{"name":"View Pull Request","url":"https://github.com/WormBase/datomic-to-catalyst/pull/128#event-1248108510"}}}

azurebrd:
Sure, it should all be working properly and ready. Thanks @a8wright

PR #130, opened 2017-09-14
Created gene_classes widget.

azurebrd:


src/rest_api/classes/laboratory/widgets/gene_classes.clj
```diff
@@ -0,0 +1,59 @@
...
+  (let [db (d/entity-db laboratory)
+        data (->> (d/q '[:find ?gene-class ?description
+                         :in $ ?laboratory
+                         :where 
+                         [?laboratoryent :gene-class.former-designating-laboratory/laboratory ?laboratory]
+                         [?gc :gene-class/former-designating-laboratory ?laboratoryent]
+                         [?gc :gene-class/id ?gene-class]
+                         [?gc :gene-class/description ?description]]
+                       db (:db/id laboratory))
+                  (map (fn [oid]
```
mgrbyte: Consider destructuring each result row rather than using `(get result index)`. 
```clojure
(map (fn [result]
            (let [[geme-class description] result]
               ....)
```

src/rest_api/classes/laboratory/widgets/gene_classes.clj
```diff
@@ -0,0 +1,59 @@
...
+                         (let [gene-class (get oid 0)
+                               description (get oid 1)]
+                           (pace-utils/vmap
```
mgrbyte: Not clear why `pace-uitils/vmap` is being used here?
Could simply use a plain map? The only case for using `vmap` here would be if you only create key/value pairs when the value is not nil (but the other keys with hard-coded values would still exist so...)
```clojure
{:description description
 :former_gene_class
 {:taxonomy "all"
  :class "gene_class"
  :label "gene-class"
  :id gene-class}}
```

mgrbyte (changes requested):
The inlined function used to transform query results into maps in  `former-gene-classes` and `gene-classes` is identical and could be factored out and given a name.

src/rest_api/classes/laboratory/widgets/gene_classes.clj
azurebrd: Good to know, thanks, fixed.

src/rest_api/classes/laboratory/widgets/gene_classes.clj
azurebrd: Good to know, didn't realize the difference between vmap and just a map.  Fixed.

src/rest_api/classes/laboratory/widgets/gene_classes.clj
azurebrd: About the main point about the query results being mapped into the same kind of hash, one tag has a keyword :former_gene_class while the other has :gene_class .  Since they're slightly different, would you prefer a single function passing [result] + a variable for the keyword being one or the other, applying (keyword) to it, and generating the map in a single function ?  I couldn't tell whether that was more readable or preferred.

mgrbyte: Having two separate functions for the two distinct endpoints is fine; I was alluding to the inline function passed to `map` in the thread-macro in both functions.
So this:
```clojure
(fn [result]
  (let [[gene-class description] result]
    {:description description
     :gene_class
     {:taxonomy "all"
      :class "gene_class"
      :label gene-class
      :id gene-class}})
```
could be factored-out and given a name e.g `(defn- process-result [result] ...)` or something like that.
Used then in both `former-gene-class` and `gene-class` end-point fns : `(map process-result)`

mgrbyte (commented):
Changes look good @azurebrd , waiting to see if you want to do the factoring as per my comment before I approve. 
:+1:

src/rest_api/classes/laboratory/widgets/gene_classes.clj
azurebrd: Ah, yes, that's what I meant.  In gene-class the map is like : 
```clojure
{:description description
     :gene_class
     {:taxonomy "all"
```
but in former-gene-class, the :gene_class tag becomes :former_gene_class  
```clojure
{:description description
     :former_gene_class
     {:taxonomy "all"
```
so it would either need two separate functions, or pass in an extra variable to change into the (keyword), and I couldn't tell whether that was better / easier to understand.

sibyl229 (commented):
Hi @azurebrd generally looks good 👍 

There is a bit of subtlety in the query that I'm not sure if it's intentional. Basically, your query would exclude gene classes that have no description. 

If it's not intentional. What you might need would be something like this (I didn't test the code below) or consider using the entity API on the laboratory entity:

```
(->> (d/q '[:find [?gc ...]
                  :in $ ?laboratory
                  :where 
                  [?gc :gene-class/designating-laboratory ?laboratory]]
          db (:db/id laboratory))
     (map (fn [gc]
              (let [gene-class (d/entity db gc)]
                {:description description
                :gene_class (pack-obj gene-class))))  ;; [rest-api.formatters.object :as obj :refer [pack-obj]
     (seq))
```

azurebrd:
Thanks @sibyl229 I didn't realize that !  I'll test that out and make changes.

azurebrd:
Thanks @sibyl229 .  I found no data that lacked a description and had either a designating laboratory or former designating laboratory, but I've changed the code to use entity on the gene-class object.  Hopefully it works, although I can't test the case until some curator creates an object that has a laboratory and lacks a description.

azurebrd:
@sibyl229 changed it to also use pack-obj.  One question to clarify, is the issue that the query requires a queried value to exist for the object to return, so it would generally be better to make subqueries if it's not clear that an object will have a value ?  I'm working on the alleles widget, and have a large query that gets from the laboratory, the alleles, the alleles's gene, method, and sequenced status, but if that means that it will only return rows with all the values, should I instead get the allele and make subqueries to follow each allele to its gene ?  I had been avoiding this because I thought it would be more efficient to make a single query, but it's not good if it will miss data doing that. Thanks :)

sibyl229 (approved):
Thanks @azurebrd ! Everything looks good.

sibyl229:
@azurebrd haha, I don't know all the answers to your question. I will tell you what I know.

I generally use the query to identify entity (ids) that is relevant, and use entity API to retrieve associated attributes. The main reason is to avoid unintentionally exclude things because of missing value, as well as to keep the query easier to read. 

In terms of performance, we know that:
* Datomic perform the query as well as Entity API functionality on the peer side (in the JVM that runs the d-2-c app)
* to do either of those tasks it relies on indexes (which are datoms sorted by certain order) found on the peer side / cache (loading from remote storage / dynamoDB if necessary)
* when Datomic loads index, it loads and caches (for future needs) not only datoms to compute the query and access data, but also nearby datoms in the sorted index
* different tasks take advantage of different indexes: http://docs.datomic.com/indexes.html
* the `:where` clauses of a query are evaluated in order, so it's good to have more specific/limiting clauses first

What I make out of this (not sure if I'm correct) is that 1) there is no fundamental difference between query calls and entity API calls in terms of performance 2) what matters (or at least a good starting point) is what/how big the indexes are being used to perform those tasks (and if the index is already found in the peer cache).

azurebrd:
Cool ! Thanks @sibyl229 =D

sibyl229:
I believe the refactoring is done as part of the changes made. Since the code between `former-gene-classes` and `gene-classes` are fairly unique to their individual cases. And the code is easy to understand. 

I will merge this so curators could test it.

azurebrd:
Thanks @sibyl229 !

mgrbyte:
nice summary @sibyl229 :arrow_up: :+1:

PR #132, opened 2017-09-26
Feature/phenotype widgets

adamjohnwright:
I have created the phenotype page widgets:

- Associated anatomy
- GO term
- RNAi
- Transgene
- Variation

I have tested these URLs by looking at the resulting JSON. They seem to be fairly simple.

After this pull request, I will make small changes based on what I see on the staging website. After that I will request curators to take a look.


src/rest_api/classes/phenotype/widgets/associated_anatomy.clj
```diff
@@ -0,0 +1,53 @@
...
+
+                    :assay
+                    (when-let [ahs (:anatomy-function/assay af)]
+                      (for [ah ahs]
+                        {:text (:ao-code/id (:anatomy-function.assay/ao-code ah))
+                         :evidence (obj/get-evidence ah)}))
+
+                    :phenotype
+                    (when-let [phs (:anatomy-function.phenotype/phenotype af)]
+                      (keys phs)
```
sibyl229: Maybe the keys need to be humanized to get rid of the namespace?

src/rest_api/classes/phenotype/widgets/variation.clj
```diff
@@ -1,8 +1,57 @@
 (ns rest-api.classes.phenotype.widgets.variation
   (:require
-   [rest-api.classes.generic-fields :as generic]))
+    [rest-api.formatters.object :as obj :refer  [pack-obj]]
+    [rest-api.classes.generic-fields :as generic]))
+
+(defn variation-info [v obs]
```
sibyl229: the `v` parameter might have been `phenotype`

src/rest_api/classes/phenotype/widgets/transgene.clj
```diff
@@ -1,8 +1,43 @@
 (ns rest-api.classes.phenotype.widgets.transgene
   (:require
+    [rest-api.formatters.object :as obj :refer  [pack-obj]]
     [rest-api.classes.generic-fields :as generic]))
 
+(defn transgene-info [t obs]
```
sibyl229: `t` parameter might have been phenotype

src/rest_api/formatters/object.clj
```diff
@@ -333,43 +333,105 @@
...
+
+   :genotype
+   (when-let [c (:anatomy-function.assay/condition holder)]
+     (first (:condition/genotype c)))
+
+   :Autonomous
+   (when-let [ss (:anatomy-function-info/autonomous holder)]
+     (for [s ss]
+       {:taxonomy "all"
+        :class "txt"
```
sibyl229, line 382: Might need to double check if the class should 'txt' or 'text'

adamjohnwright: This is how the object are coming from Catalyst API. I am leaving it the same for now. In the future we could look into making this more consistent.

PR #135, opened 2017-09-29
Alleles + Strains widgets for Laboratory class.

azurebrd:


src/rest_api/classes/laboratory/widgets/alleles.clj
```diff
@@ -0,0 +1,42 @@
...
+        data (->> (d/q '[:find [?varent ...]
+                         :in $ ?laboratory
+                         :where 
+                         [?laboratory :laboratory/alleles ?varent] ]
+                       db (:db/id laboratory))
+                  (map (fn [wbvar]
+                         (let [allele (d/entity db wbvar)]
+                           (pace-utils/vmap
+                             :sequenced 
+                             (if (= (str (:variation/seqstatus allele))
```
mgrbyte: No need to cast to `str` here for equality testing, can simply do:
```clojure
(if (= (:variation/seqstatus allele) :variation.seqstatus/sequenced)
   "yes"
   "no")
```

src/rest_api/classes/laboratory/widgets/alleles.clj
```diff
@@ -0,0 +1,42 @@
...
+                                    ":variation.seqstatus/sequenced") "yes" "no")
+                             :type 
+                             (if (:method/id (:locatable/method allele))
+                               (-> (:method/id (:locatable/method allele))
+                                   (str/replace #"_allele" "")
+                                   (str/replace #"Allele" "")))
```
mgrbyte: No problem here, but could use `some->` to be more succinct:
```clojure
:type (some-> (:method/id (:locatable/method allele))
              (str/replace #"_allele" "")
              (str/replace #"Allele" ""))
```

mgrbyte (changes requested):
Generally looks good.
One change and one suggestion :+1:

adamjohnwright (approved):

adamjohnwright:
I agree with Matts suggestions. I don't see anything else, in that way of style, that needs to change. Great work

src/rest_api/classes/laboratory/widgets/alleles.clj
azurebrd: Ah, cool, thanks.

src/rest_api/classes/laboratory/widgets/alleles.clj
azurebrd: Oooh, that's really neat, thanks Matt !  I've made the changes, I appreciate the feedback while travelling =)

src/rest_api/classes/laboratory/widgets/alleles.clj
```diff
@@ -16,13 +16,11 @@
                          (let [allele (d/entity db wbvar)]
                            (pace-utils/vmap
                              :sequenced 
-                             (if (= (str (:variation/seqstatus allele))
+                             (if (= (:variation/seqstatus allele)
                                     ":variation.seqstatus/sequenced") "yes" "no")
```
mgrbyte: Value being compared shouldn't be a string (is it?) (i.e: ":variation.seqstatus/sequenced" _should_ be a keyword/ident not a string)

azurebrd:
@mgrbyte Right !  Thanks for catching that, it seemed odd that it didn't need a (str) casting earlier, and I didn't realize your suggestion was comparing it straight to a keyword/ident.  Thanks !

mgrbyte (approved):

PR #136, opened 2017-10-03
Added and tested sequence feature widgets for cds, operon, pseudogene…

adamjohnwright:
Added and tested sequence feature widgets for:
- cds
- operon
- pseudogene
- transcript
- transposon

There is only one Operon with associated features (CEOP3666). I feel like we could either include a widget or not for the Operon page. I am leaning toward adding it. Either way we could have a REST endpoint for this widget. 

Also to node is this time I took on more the style that you ( @sibyl229 ) suggested - it is a good fit. Where we don't dynamically generate the keyword. 

I have switched port 3000 to be running this branch. I have gone through each of the widgets and it appears as if they are all running correctly!

### DIFF
--- a/src/rest_api/classes/transcript/widgets/feature.clj
+++ b/src/rest_api/classes/transcript/widgets/feature.clj
@@ -6,18 +6,17 @@
 
 (defn associated-features [transcript]
   (let [db (d/entity-db transcript)
-        data  (->>
-                (d/q '[:find [?f ...]
-                       :in $ ?transcript
-                       :where
-                       [?fg :feature.associated-with-transcript/transcript ?transcript]
-                       [?f :feature/associated-with-transcript ?fg]]
-                   db (:db/id transcript))
-                (map (partial feature/associated-feature db))
-                (seq))]
-    {:data (not-empty data)
+        data (->> (d/q '[:find [?f ...]
+                         :in $ ?transcript
+                         :where
+                         [?fg :feature.associated-with-transcript/transcript ?transcript]
+                         [?f :feature/associated-with-transcript ?fg]]
+                       db (:db/id transcript))
+                  (map (partial feature/associated-feature db))
+                  (seq))]
+    {:data data
      :description "Features associated with this transcript"}))
 
 (def widget
-    {:name generic/name-field
-     :features associated-features})
+  {:name generic/name-field
+   :features associated-features})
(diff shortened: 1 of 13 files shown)


src/rest_api/classes/transposon/widgets/feature.clj
```diff
@@ -0,0 +1,23 @@
...
+        data  (->>
+                (d/q '[:find [?f ...]
+                       :in $ ?transposon
+                       :where
+                       [?fg :feature.associated-with-transposon/transposon ?transposon]
+                       [?f :feature/associated-with-transposon ?fg]]
+                   db (:db/id transposon))
+                (map (partial feature/associated-feature db))
+                (seq))]
+    {:data (not-empty data)
```
mgrbyte: No need for `not-empty` since `seq` jas been applied already

src/rest_api/classes/transcript/widgets/feature.clj
```diff
@@ -0,0 +1,23 @@
+(ns rest-api.classes.transcript.widgets.feature
+  (:require
+    [datomic.api :as d]
+    [rest-api.classes.feature.core :as feature]
+    [rest-api.classes.generic-fields :as generic]))
+
+(defn associated-features [transcript]
+  (let [db (d/entity-db transcript)
+        data  (->>
```
mgrbyte: think it's best to always have an expression at the start of a threaded macro, since to my eyes it looks like something is missing (!) e.g:

```clojure
(->> something
     something-else)
```

src/rest_api/classes/transposon/widgets/feature.clj
```diff
@@ -0,0 +1,23 @@
+(ns rest-api.classes.transposon.widgets.feature
+  (:require
+    [datomic.api :as d]
+    [rest-api.classes.feature.core :as feature]
+    [rest-api.classes.generic-fields :as generic]))
+
+(defn associated-features [transposon]
+  (let [db (d/entity-db transposon)
+        data  (->>
```
mgrbyte: Same as threaded macro comment above.

mgrbyte (changes requested):
Looks great @a8wright :+1: 
Just a couple of minor style adjustments (if you agree)

src/rest_api/classes/transcript/widgets/feature.clj
adamjohnwright: good idea. I see what you are saying

src/rest_api/classes/transposon/widgets/feature.clj
adamjohnwright: I just tested this and is seams to be true.

mgrbyte (approved):
:+1:

PR #137, opened 2017-10-03
Adding history wigets for protein, paper and feature pages

adamjohnwright:
Created history widgets on the following pages:

- Protein
- Paper
- Features

This has been tested for all of the widgets. Unfortunately they are not too similar. This will require curator testing for sure in order to make sure an edge case has not been missed

### DIFF
--- a/src/rest_api/classes/feature.clj
+++ b/src/rest_api/classes/feature.clj
@@ -1,6 +1,7 @@
 (ns rest-api.classes.feature
   (:require
     [rest-api.classes.feature.widgets.overview :as overview]
+    [rest-api.classes.feature.widgets.history :as history]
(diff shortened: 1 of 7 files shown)


mgrbyte (changes requested):
Generally looks fine :+1: 
Minor style change request: pls use single [space indent][1] for function calls.

[1]: https://github.com/bbatsov/clojure-style-guide#one-space-indent

mgrbyte (approved):
Better than before, but it seems the vim plugin is doing it wrong.
Function calls that have arguments spanning several lines:should only have a one space indent (style guide) 
```clojure
(fn
 calls
 like-this
 should-have-a-single-space-indent)
```
But I don't think I should/can block this PR because of this (no effect on running code - perhaps something you could look into)

PR #147, opened 2017-12-07
return 200 even if id convert for blast hit fail to find match

sibyl229:
Hi @a8wright this should help with some of the 404s that are affecting EB health check.


sibyl229:
Thanks @a8wright !

PR #149, opened 2017-12-08
Feature/262 interaction

sibyl229:
Some hack-ish logic for interaction type here:
- filter out legacy types 
- reformat type string
- for the table and the graph, decide which type to show when multiple types are present 
- for interaction overview, show all applicable types.


mgrbyte (approved):
:+1:

sibyl229:
Thanks a lot @mgrbyte !

PR #150, opened 2017-12-11
Feature/laboratory

azurebrd:
Lab Members widget.


src/rest_api/classes/laboratory/widgets/members.clj
```diff
@@ -1,31 +1,108 @@
...
+        lineage (generate-lineage laboratory)
+        data (->> (d/q '[:find [?members ...]
+                         :in $ ?laboratory
+                         :where
+                         [?laboratory :laboratory/past-lab-members ?members]]
+                       db (:db/id laboratory))
+                  (map (fn [members]
+                         (let [member (d/entity db members)
+                               lab (d/entity db laboratory)]
+                           {:name (pack-obj "person" member)
```
adamjohnwright: You don't need to have "person" in the middle. pack-obj should be able to figure out the entity type on its own.

src/rest_api/classes/laboratory/widgets/members.clj
```diff
@@ -1,31 +1,108 @@
...
+    lineage))
+
+(defn current-members [laboratory]
+  (let [db (d/entity-db laboratory)
+        lineage (generate-lineage laboratory)
+        data (->> (d/q '[:find [?members ...]
+                         :in $ ?laboratory
+                         :where
+                         [?laboratory :laboratory/registered-lab-members ?members]]
+                       db (:db/id laboratory))
```
adamjohnwright, line 69: i am pretty sure you could just do (:laboratory/_registered-lab-members laboratory). This will give you the reverse relationship that you are looking for.

src/rest_api/classes/laboratory/widgets/members.clj
```diff
@@ -1,31 +1,108 @@
...
+               (flatten 
+                 (for [rep reps] 
+                   (->> (d/q '[:find ?other-person-id ?other-person-name (pull ?person-supervised-by-ent [*])
+                               :in $ ?person
+                               :where
+                               [?other-person-ent :person/supervised-by ?person-supervised-by-ent]
+                               [?person-supervised-by-ent :person.supervised-by/person ?person]
+                               [?other-person-ent :person/id ?other-person-id]
+                               [?other-person-ent :person/standard-name ?other-person-name]]
+                             db (:db/id rep))
```
adamjohnwright, line 57: This could be something like:

 (some->> (:laboratory/representative laboratory)
                 (map :person.supervised-by/_person)
                 (map :person/_supervised-by))

I am not sure what you are doing in the find part of the query. But I am pretty sure it the full query could be avoided through something similar to the code i am suggesting.

src/rest_api/classes/laboratory/widgets/members.clj
```diff
@@ -1,31 +1,108 @@
...
+                        (map (fn [oid]
+                               (generate-map oid)))(seq)))))]
```
adamjohnwright: This could be simplified as (map generate-map) instead of (map (fn [old] (generate-map old)))

src/rest_api/classes/laboratory/widgets/members.clj
```diff
@@ -1,31 +1,108 @@
...
+        other-person-name (get oid 1)
+        json-data (get oid 2)
+        roles (filter #(re-find #"role" (str %)) (keys json-data))
+        level
+        (for [role roles
+              :let
+              [level (-> role
+                         (str/replace #":role/" "")
+                         (str/replace #"-" " ")
+                         (str/capitalize))]] level)
```
adamjohnwright: This part could be:
(some->> roles
                (map name)
                (map obj/humanize-ident))

src/rest_api/classes/laboratory/widgets/members.clj
```diff
@@ -1,31 +1,108 @@
...
 
-(defn former-members [l]
-  {:data (some->> (:labloratory/past-lab-members l)
-                  (map get-lineage-data))
-   :description "former members of the laboratory"})
+
+(defn- generate-lineage [laboratory]
+  (let [db (d/entity-db laboratory)
+        reps (:laboratory/representative laboratory)
+        lineage
```
adamjohnwright, line 46: I think that your code would simplify a lot if you were to remove extra variables like "lineage" here. Instead of defining lineage in the let an then returning lineage inside the let. The logic for creating lineage should just be in side the let. After my next suggestion you should be able to get rid of the let all together. But it would be good to take out these extra variables. Once they are removed I think i could give much better suggestions on how to simplify the code.

src/rest_api/classes/laboratory/widgets/members.clj
```diff
@@ -1,31 +1,108 @@
...
+  (let [db (d/entity-db laboratory)
+        lineage (generate-lineage laboratory)
+        data (->> (d/q '[:find [?members ...]
+                         :in $ ?laboratory
+                         :where
+                         [?laboratory :laboratory/registered-lab-members ?members]]
+                       db (:db/id laboratory))
+                  (map (fn [members]
+                         (let [member (d/entity db members)
+                               lab (d/entity db laboratory)]
```
adamjohnwright: I don't think you are using the lab variable anywhere.

src/rest_api/classes/laboratory/widgets/members.clj
```diff
@@ -1,31 +1,108 @@
...
+                           {:name (pack-obj "person" member)
+                            :level (:level ((keyword (:person/id member)) lineage))
```
adamjohnwright: I am not sure what ((keyword (:person/id member)) lineage)) does but you could pull it out into a variable instead of regenerating it for each of these lines.

src/rest_api/classes/laboratory/widgets/members.clj
```diff
@@ -1,31 +1,108 @@
...
+        data (->> (d/q '[:find [?members ...]
+                         :in $ ?laboratory
+                         :where
+                         [?laboratory :laboratory/past-lab-members ?members]]
+                       db (:db/id laboratory))
+                  (map (fn [members]
+                         (let [member (d/entity db members)
+                               lab (d/entity db laboratory)]
+                           {:name (pack-obj "person" member)
+                            :level (:level ((keyword (:person/id member)) lineage))
```
adamjohnwright: This is same as above. You could pull ((keyword (:person/id member)) lineage)) into it's own variable

adamjohnwright (changes requested):

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: Thanks, I see that works, and it's good to know about obj/humanize-ident .
I see that it works without the (map name), (name) is the first thing applied by obj/humanize-ident .

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: Right, thanks !

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: Cool, thanks !

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: I either don't understand the advantage that, or don't understand what you mean.  You mean in the let to assign that to its own variable ?  Would that be just for clarity, it seems like an extra variable.

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: Oh, I see, I didn't think it would be computationally expensive, compared to creating extra variables.  Sure though, changed.

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: Changed.

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: I couldn't figure out how to pull out the role from ?person-supervised-by-ent without using a (pull) like [:find ?other-person-id ?other-person-name (pull ?person-supervised-by-ent [*]) which (generate-map) then uses the json-data to extract it.

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: Oh, cool, thanks !

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: That makes sense, but         members (:laboratory/_registered-lab-members laboratory)
then putting  :members members in the { } output gives me  "members" : null  in the .json output.

src/rest_api/classes/laboratory/widgets/members.clj
azurebrd: I'm not sure how you mean, isn't the logic for lineage inside the (let) already ?  I wanted to pass 'registered-lab-members' and 'past-lab-members' as variables for the query, so that former-members and current-members could be calling the same function passing either string, then the function defining lineage could be written once inside that new function, but I couldn't figure out the syntax for that, and got this to work.

adamjohnwright (approved):
This looks good. Thanks!

azurebrd:
Oh, awesome, thanks @a8wright !

Issue #151, opened 2017-12-12
lineage 500 error


adamjohnwright:
I found this in the nginx production log: 

52.23.219.146 - - [12/Dec/2017:12:49:02 +0000] "GET /rest/widget/person/WBPerson635/lineage HTTP/1.1" 500 67 "-" "HTTP-Tiny/0.034"

I just tested it on staging and it caused a gateway timeout 

I think there probably be work done to figure why it is using so much resources.

azurebrd:
Thanks @a8wright .  I'm looking into it, but can't tell how to untangle where to start.

Looking at juancarlos.wormbase widgets worked for other classes, but no widgets worked for person (e.g. overview).  I switched the develop branch and did a 'git pull' and kept getting the same errors.  I tried to stop and restart the datomic-to-catalyst server : lein ring server-headless 4001   but it's getting an error : 
Exception in thread "main" java.lang.RuntimeException: No such namespace: generic, compiling:(rest_api/classes/laboratory/widgets/members.clj:101:1)

I don't know whether something didn't go properly into develop, and that's why it can't start, and I'm confused as to why the person widgets didn't work, but other widgets did.  Do you have some advice on how to get my copy of datomic-to-catalyst to a point that runs so I can branch it and debug this error (which I do see on staging as having other person widgets working).  Thanks !

azurebrd:
@a8wright Oh, I think I see why that's happening.  Will need to patch the laboratory class members.clj  to account for an extra data field when merging my code with your stubs.  There's still another error past that though : 

Started nREPL server on port 41460
Exception in thread "main" java.net.BindException: Address already in use, compiling:(/tmp/form-init3015458591603639907.clj:1:73)

I have no idea about that.

azurebrd:
@a8wright Found the problem.  I can start debugging it now.  Thanks for merging the other bit with members.clj

mgrbyte:
Likely cause is the use of `*` in pull expressions [here][1], [here][2] and [here][3].

This will select _all_ attributes in the relation(s); better to only pull the attributes that are needed.
The attributes select should include just those needed by the `:where` clause of the query, and the `generate-map` function (since the result of these queries are fed into that function).

Rationale for the above hypothesis is that using [wildcarding][5] in datomic pull expressions will pull all component attributes recursively, and `supervised-by` relation of a person is such an attribute (schema marked up with `:db/isComponent`). 
[...]

PR #153, opened 2017-12-18
Hotfix/0.2.4.1 - issue #151

adamjohnwright:
This hotfix removes the datomic queries that were likely causing issues with generating the person lineage graphs. 

There are two things left to do (as far as I am aware):
- Add code that makes the size of the nodes relative to the largest. 
- Fix the finctions the call them selves so that they only recur n times up or down the tree.


src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -2,378 +2,298 @@
...
+           (map (fn [role]
+            (obj/humanize-ident
+              (name role))))))
+
+(defn- get-duration [holder]
+  (some->> (get-roles holder)
+           (map
+             (fn [role-key]
+               (let [rfrom (keyword
+                             (-> role-key
```
mgrbyte: Consider pulling out the role keyword generation into a fn.
e.g: in the let
```clojure
   (let [role-keyword (partial keyword (-> role-key 
                                           (str/replace #"/" ".") 
                                           (str/replace #":" "")))
         rfrom (role-keyword "from")
         rto (role-keyword "to")]
          ...)
```

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -2,378 +2,298 @@
...
+                    (:person.supervised-by/person holder))))
+           (map
+             (fn [supervisor]
+               {:source (str direct-or-full (:id (:name supervisor)))
+                :target (str direct-or-full (:person/id person))
+                :role (first (:level supervisor))
+                :targetArrowShape "triangle"
+                :lineStyle (if (= direct-or-full "Direct") "dashed" "solid")
+                :lineColor (get role-colour (first (:level supervisor)))}))))
+
```
mgrbyte: As we've discussed elsewhere, depending on the depth of this graph-like relationship, 
these mutally recursive function calls can/will cause a StackOverflow exception.

I think for these (matually) recursive functions you might want to look into using `trampoline`.
The mutual recursion is between the outer `defn recursive-*` function and the anonymous function invoked within the map expression in each function.
I believe these functions could be tweaked to use [trampoline][1] to avoid the StackOverflow issue; that being said I'd have to experiment as not used it before.

[1]: https://clojuredocs.org/clojure.core/trampoline

mgrbyte (commented):
Generally looks much cleaner.
You're probably most interested in the trampoline stuff I commented on inline below.
Wondering if you're running into StackOverflow with any of your testing with this?

src/rest_api/classes/person/widgets/lineage.clj
mgrbyte: If the recursion level is being hit, consider using a datalog query with (recursive) rules, which may well produce a similar solution.

adamjohnwright: I did not know about trampoline. I think this is a great idea.

adamjohnwright:
@azurebrd I finally got a chance to finish off this widget. I am sure that it is mainly correct. I just don't know if it works exactly like you intend it to. In particular it is not producing any errors when run. Including for WBPerson77. One area that I am not sure of is if the scale is correctly calculated. I am using the largest scale node value to calculate the radius of the nodes. 

Could you take over this branch and make sure it does what you think it ideally should?

I can give you a walk through on Skype if you like. I think it would be best for you to have a chance to go through the code on your own before we go through it.

Keep in mind I have not tested it with the UI. I am guessing it will need a little work before we do that. 

@mgrbyte I used your suggestion from months ago to make the duration function more concise. 

A

chris-grove:
@a8wright 
FYI, @azurebrd is on vacation until next Monday, just so you know

src/rest_api/classes/person/widgets/lineage.clj
adamjohnwright: This is incorporated in.

azurebrd:
Thanks @chris-grove !

Thanks @a8wright !  I'm just back from traveling and will look at Caltech things that may be pressing, then take a look at this as soon as I get a chance.  Possibly will also talk to Raymond, as he was very involved in deciding how the logic should work.

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -2,378 +2,309 @@
...
-                                       (conj (filter-other-nodes "Full" largest-node-scaling-full queried-data-recurse-supervisors) (filter-this-node "Full" queried-data-this-person))))))" } } ]")
-        elements-full (str "{ " nodes-full ", " edges-full " }")
-        elements-direct (str "{ " nodes-direct ", " edges-direct " }")
-        ]
-    {:existingRolesFull existing-roles-full
-     :existingRolesDirect existing-roles-direct
-     :thisPerson this-person-id 
-     :elementsFull (if elements-full elements-full)
-     :elementsDirect (if elements-direct elements-direct)
+  (let[{nodes :nodes
```
mgrbyte: add a space

src/rest_api/classes/person/widgets/lineage.clj
```diff
@@ -2,378 +2,309 @@
...
+         roles :roles
+         edges :edges} (get-graph person)]
+    {:existingRolesFull roles
+     :existingRolesDirect (roles-direct person)
+     :thisPerson (:person/id person)
+     :elementsFull(when-let [elements (elements-full nodes edges)]
```
mgrbyte, line 279: add a space

mgrbyte (commented):
Two very minor formatting things,
overall the code looks great :+1:

azurebrd:
Hi @a8wright , I'm just now getting a chance to look at the changes.  It looks a lot better !

It's interesting to see how the person-supervised entity can get passed to a function, and the role duration and level extracted from there, instead of having to query with a pull [*] .  Also seeing how you flag whether a get-supervisor-graph has been visited.

As far as I can tell, then largest node scaling in elements-full makes sense, but I don't see it in elements-direct.  I can't tell where the radius is getting set for supervisors/supervisees , should it be generated at get-node ?

Just looking at the code, everything seems to make sense, but I don't know how the format of the output is.  Thanks Adam.

adamjohnwright:
Hi @azurebrd I am glad the full map makes sense to you. I just spent a little time trying to remember how the direct edges works and I am not sure if it is currently working correctly with scaling. I think you are right that it should use the same scaling as the full graph - with get-node. Scaling is one of the last things that I figured out so I am not sure if it was fully checked over with the direct nodes. Let me know what you find out. 

As a side note I merged this branch in and then commented out the endpoint on staging. This was just to make sure it didn't diverge from the main code base too much. Feel free to create a new branch to test your changes on.

azurebrd:
Cool, @a8wright , I'll check out the code and try it out when I get a chance.

PR #157, opened 2018-01-29
Feature/corresponding all widgets

adamjohnwright:
This was full checked by curation a while ago. I think this might have been held back because it was requiring template changes for the sequence widget on the gene page. And maybe other sequence widgets. 

I think what we should do is pull it in and then look at the various pages that have sequence widgets.


adamjohnwright:
It goes with this issue: https://github.com/WormBase/website/issues/5773

src/rest_api/classes/protein/widgets/overview.clj
```diff
@@ -25,38 +25,15 @@
...
-              :length_spliced nil
-              :type (when-let [cds (:cds/_corresponding-protein cph)]
-                      (if-let [type-field (:method/id (:locatable/method cds))]
-                        (str/replace type-field #"_" " ")))}))
-   :description "corresponding cds, transcripts, gene for this protein"})
-
 (def widget
   {:name generic/name-field
-   :estimated_molecular_weight estimated-molecular-weight
+ ;  :estimated_molecular_weight estimated-molecular-weight
```
sibyl229: Might need to uncomment

src/rest_api/classes/protein/widgets/overview.clj
```diff
@@ -25,38 +25,15 @@
...
-   :remarks generic/remarks
+ ;  :pfam_graph pfam-graph
+ ;  :best_human_match best-human-match
+ ;  :taxonomy generic/taxonomy
+ ;  :description generic/description
+ ;  :estimated_isoelectric_point estimated-isoelectric-point
+ ;  :remarks generic/remarks
    :type type-field
-   :corresponding_all corresponding-all})
+   :corresponding_all generic/corresponding-all})
```
sibyl229, line 39: There are some fields that aren't ported over to Datomic

PR #160, opened 2018-02-17
Feature/variation page

adamjohnwright:
This includes changes for the variation page's widgets isolation, genetics and molecular details. Where isolation and genetics are ready for curator review. And molecular details needs a lot of work (big rabbit hole) and requires sequences. 

These were checked by using a lot of different URLs - fields are sparsely populated.


src/rest_api/classes/generic_functions.clj
```diff
@@ -22,3 +24,44 @@
...
+(defn- strain-list  [strains]
+  (seq  (map  (fn  [strain]
+                (let  [tgs  (:transgene/_strain strain)]
+                  (pace-utils/vassoc
+                    (pack-obj "strain" strain)
+                    :genotype  (:strain/genotype strain)
+                    :transgenes  (pack-obj "transgene"  (first tgs)))))
+             strains)))
+
+(defn categorize-strains [strains]
```
mgrbyte: Is it worth abstracting a  function predicate here?
e.g:
```
(defn include-strain? [cgc-pred]
  #(and (not (seq (:transgene/_strain %)))
        (= (count (:gene/_strain %)) 1)
        (cgc-pred %))
;; used like:
(->> strains (filter (include-strain? is-cgc?)) strain-list)
;; and
(->> strains (filter (include-strain? (complement is-cgc?)) strain-list)

```

src/rest_api/classes/variation.clj
```diff
@@ -1,6 +1,9 @@
 (ns rest-api.classes.variation
   (:require
     [rest-api.classes.variation.widgets.overview :as overview]
+    [rest-api.classes.variation.widgets.genetics :as genetics]
+    [rest-api.classes.variation.widgets.isolation :as isolation]
+    ;[rest-api.classes.variation.widgets.molecular-details :as molecular-details]
```
mgrbyte: Perhaps a comment as to why molecular-details is commented (pending figuring out where to get the data from/some other issue?)

src/rest_api/classes/variation/widgets/genetics.clj
```diff
@@ -0,0 +1,88 @@
...
+           [genes])
+   :description "gene in which this variation is found (if any)"})
+
+(defn reference-allele [variation]
+  {:data (some->> (:gene.reference-allele/_variation variation)
+                  (map :gene/_reference-allele)
+                  (map pack-obj))
+   :description "the gene that has this variation as a reference allele"})
+
+(defn other-allele [variation]
```
mgrbyte, line 28: General note on use of `not-empty` with `some->`/`some->>`:
`not-empty` can be placed as the last expression in the macro here, since `some->`/`some->>` will return nil if one of the expressions evaluates to an empty sequence/nil.
No logical difference, might make the code more readable.

src/rest_api/classes/variation/widgets/molecular_details.clj
```diff
@@ -0,0 +1,315 @@
...
+;has data
+(defn context [variation]
+  {:data (let [public-name (:variation/public-name variation)
+               max-seqlen 1000000
+               flank 250
+               seq-len (if
+                         (contains? variation :variation-source-location)
+                          nil
+                          1)
+                             ]
```
mgrbyte: formatting

src/rest_api/classes/variation/widgets/molecular_details.clj
```diff
@@ -0,0 +1,315 @@
...
+                             "Transposon insertion"
+
+                             (contains? variation :variation/transposon-excision)
+                             "Transposon excision"
+
+                             :else
+                             (->> type-of-mutation-map
+                                  (map
+                                    #(str (if
+                                            (contains? variation (first %))
```
mgrbyte, line 77: minor nit-pick; spacing issue here (anon fn should line up under the m of "map")

src/rest_api/classes/variation/widgets/molecular_details.clj
```diff
@@ -0,0 +1,315 @@
...
+       "..."
+       (subs peptide (- position 20) 19)
+       " "
+       "<b>" (subs peptide (- position 1) 1) "</b>"
+       " "
+       (subs peptide position 20)
+       "..."
+       (+ position 19)))
+
+(defn- do_markup [peptide var-start variation mutation-type]
```
mgrbyte: Not sure why this stuff isn't done in client-side js, I guess the perl code did, so you've just re-produced?

src/rest_api/classes/variation/widgets/molecular_details.clj
```diff
@@ -0,0 +1,315 @@
...
+                            "background-color:yellow;font-weight:bold;text-transform:uppercase;"
+                            "background-color:yellow")
+                   :deletion_with_insertion "background-color: #FF8080; text-transform:uppercase;"}
+       var-stop (if (= (count variation) 0)
+                  (+ var-start 1)
+                  (+ var-start (count variation)))
+       sequence (if (= (count variation) 0)
+                  (str (subs peptide 0 var-start) "-" (subs peptide (+ var-start (count variation))))
+                  peptide)
+                 ]                                                     )
```
mgrbyte: Seems something went awry with formatting at the end of this fn (do_markup)

src/rest_api/classes/variation/widgets/molecular_details.clj
mgrbyte: probably you meant `do-markup` not `do_markup`

mgrbyte (changes requested):
Other than formatting issues, and some slight duplication, looks ok. Is there anything specific you want me to test/look out for here?

src/rest_api/classes/variation/widgets/molecular_details.clj
adamjohnwright: Yes, especially this molecular details widget turned into a rabbit hole. I think the main reason is because this logic needs to be applied to the data before the perl template works on it. If we were to start using React js for rendering the widgets we would be able to do a lot more on the client side with regards to data manipulation. For now the easiest thing to do is to continue replicating the catalyst functionality. we don't have too much longer till these endpoints are fixed and then we can start discussing other architectural things.

src/rest_api/classes/generic_functions.clj
adamjohnwright: Not too sure what you mean with this one. I will have to discuss this with you on slack

src/rest_api/classes/variation/widgets/genetics.clj
adamjohnwright: thanks

src/rest_api/classes/generic_functions.clj
```diff
@@ -22,3 +24,44 @@
 
     (contains? h :qualifier/partial)
     "Partial"))
+
+(defn- is-cgc?  [strain]
+  (some #(=  (->>  (:strain.location/laboratory %)
```
sibyl229: There are several places where you used double space. Is this intentional?

sibyl229 (approved):

src/rest_api/classes/generic_functions.clj
adamjohnwright: This is not intentional but I know what causes it. If I cur and paste code into vim it will try to format it and for some reason changes it to double space. If I turn off the formatting and then paste it doesn't to id. I must have pasted this in from somewhere else without turning off vims formatting.

PR #164, opened 2018-03-06
Adding expression pattern widgets for antibody, transgene and construct

adamjohnwright:
This is simply the table that was made for life stage page expression widget. There are a couple features to make it slightly more complete and has added expression widgets for antibody, transgene and construct

### DIFF
--- a/src/rest_api/classes/expr_pattern/core.clj
+++ b/src/rest_api/classes/expr_pattern/core.clj
@@ -6,14 +6,25 @@
 
 (defn pack [ep ep-holder]
   (let [h (first (:expr-pattern/gene ep))
-	gene (:expr-pattern.gene/gene h)]
+        gene (:expr-pattern.gene/gene h)]
     {:certainty (generic-functions/certainty ep-holder)
      :expression_pattern (pack-obj ep)
      :reference (some->> (:expr-pattern/reference ep)
-			 (map :expr-pattern.reference/paper)
+                         (map :expr-pattern.reference/paper)
                          (map pack-obj))
      :gene (some->> (:expr-pattern/gene ep)
                     (map :expr-pattern.gene/gene)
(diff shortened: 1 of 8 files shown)


mgrbyte (approved):
indent is wonky in anon functions e.g line 17), but other than that :+1:

src/rest_api/classes/expr_pattern/core.clj
```diff
@@ -6,14 +6,25 @@
...
+                    (or
+                      (some->> (:expr-pattern/subcellular-localization ep)
+                               (sort)
+                               (str/join "<br />")
+                               (not-empty))
+                      (some->> (:expr-pattern/remark ep)
+                               (map :expr-pattern.remark/text)
+                               (sort)
+                               (str/join "<br />")
+                               (not-empty))))}))
```
sibyl229: Not sure if `(not-empty)` works on a string. It takes a collection as an input. But if you think it's fine, then it's okay.

src/rest_api/classes/transgene/widgets/expression.clj
```diff
@@ -0,0 +1,24 @@
...
+
+(defn marker-for [t]
+  {:data (some->> (:transgene/marker-for t)
+                  (first)
+                  (:transgene.marker-for/text))
+   :description "string decribing what the transgene is a marker for"})
+
+(defn expression-patterns [t]
+  {:data (some->> (:expr-pattern/_transgene t)
+                  (map (fn [ep]
```
sibyl229, line 17: `map` returns an array, possibly empty one, and you probably need to use `seq` to convert it to `nil`

mgrbyte: Is fine because `some->>` will return nil if the last expression is empty or nil.

PR #171, opened 2018-04-26
Fixes lack of dashed lines in supervisor edges in direct graph.

azurebrd:
Finds error for lack of scaling in Direct graph nodes, keeps lack of
scaling for Full graph nodes.

### DIFF
--- a/src/rest_api/classes/person/widgets/lineage.clj
+++ b/src/rest_api/classes/person/widgets/lineage.clj
@@ -194,7 +194,7 @@
                      (into {}))
      :edges (conj supervisor-edges supervisee-edges)}))
 
-(defn- scale-nodes-map [nodes direct-or-full]
+(defn- scale-nodes-map [nodes]
   (when (some? nodes)
     (let [largest-node-scaling (some->> nodes
                                         (map :scaling)
@@ -210,7 +210,7 @@
                                 (conj node {:radius radius}))]
                      {:data data})))))))
 
-(defn- elements [nodes-map edges-map direct-or-full]
+(defn- elements [nodes-map edges-map]
   (pace-utils/vmap
     :edges
     (some->> (vals edges-map)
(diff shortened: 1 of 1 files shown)


azurebrd:
Line 205 doesn't work, but is left as a placeholder with a potential fix.

Line 206 doesn't work because :radius already exists, so (conj) doesn't replace the old value with the new one.  I think (update-in) should be able to replace the value, but I can't figure out the syntax for replacing the existing value with the radius.  

Added some code to only call scale-nodes-map when it's "Direct", but still needs to add a restriction to not call it when "is-primary"

adamjohnwright:
Looks like good progress. Could you try conj with the elements in the other order? I think the first gets replaced with the second one.

azurebrd:
Ah, thanks, that works.  I see other problems with it now, the Full view doesn't get updated.  I'll work on that.

adamjohnwright (approved):

azurebrd:
I've updated the code to fix the scaling of the primary node in direct view, and the error with loading the full view.  Still needs testing of other objects, will keep you updated.

azurebrd:
Updated Full view to keep scaling.  Finding an error with WBPerson77, will debug.

azurebrd:
Error also exists in staging version. Browser console says : 

SyntaxError: missing } after property list[Learn More]
jquery.js:9597:4
note: { opened at line 3, column 301450
jquery.js:9597:4

Works on live site, but can't tell how the json output is different.

azurebrd:
Found and fixed some more errors:
Persons with standard name blank now have a default.
Supervise graph direction with no edges now get the nil role filtered out.
Persons with apostrophe in their names have them removed, but this is not ideal.  @a8wright do you how we could make the syntax to end up with \\' in the output ?  Line 243 has an example of what it should look like, and that should be fixed in line 62.  I tried \' \\' (results in just ') \\\\' (results in \\\\') other amounts of \ give errors and don't generate data.

PR #173, opened 2018-05-30
making it pull the sequence for Fosmid and can be used for others

adamjohnwright:
This has been tested with this endpoint: curl -H content-type:application/json http://localhost:3000/rest/widget/clone/WRM0636cG07/sequences | jq '.'

Once we get it pulled in we can use the functions for other sequences as well. We can also get CG to test it once we have the template changed. 

@sibyl229 the object for the sequence being returned has the header field as "Sequence" for when the actual clone sequence is known and "Reference Sequence" for when it is pulled from the refseqdb. Feel free to suggest other JSON patterns if this does not work well with the template.

### DIFF
--- a/src/rest_api/classes/clone/widgets/sequences.clj
+++ b/src/rest_api/classes/clone/widgets/sequences.clj
@@ -2,6 +2,7 @@
   (:require
     [clojure.string :as str]
     [rest-api.classes.generic-fields :as generic]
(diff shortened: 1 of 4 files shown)


sibyl229 (approved):

sibyl229:
@a8wright Please double check if you need to compute reverse complementary for (-) strand

PR #174, opened 2018-06-01
265 interaction

sibyl229:
The objective is to reuse the gene page's interaction widget on wbprocess page and interaction page.

Technical summary:
- The concept of focus gene isn't applicable on other pages. Functions that rely on the focus gene are re-written to avoid focus gene
- Rewrite queries to get the set of correct tuples (`interaction-interactorholder1-interactorholder2`) at query time to avoid subsequent filtering (that is somewhat difficult to track).
- Add interactions widget for wbprocess page and interaction page
- Support string-based "other interactors"

### DIFF
--- a/src/rest_api/classes/interaction/core.clj
+++ b/src/rest_api/classes/interaction/core.clj
@@ -46,38 +46,38 @@
   (some-fn :molecule/id :gene/id :rearrangement/id :feature/id))
 
 (def int-rules
-  '[[(x->interaction ?gene ?h ?int)
+  '[[(->interaction ?gene ?h ?int)
      [?h :interaction.interactor-overlapping-gene/gene ?gene]
      [?int :interaction/interactor-overlapping-gene ?h]]
-    [(x->interaction ?feature ?h ?int)
+    [(->interaction ?feature ?h ?int)
      [?h :interaction.feature-interactor/feature ?feature]
      [?int :interaction/feature-interactor ?h]]
-    [(x->interaction ?molecule ?h ?int)
+    [(->interaction ?molecule ?h ?int)
      [?h :interaction.molecule-interactor/molecule ?molecule]
      [?int :interaction/molecule-interactor ?h]]
-    [(x->interaction ?rearrangement ?h ?int)
+    [(->interaction ?rearrangement ?h ?int)
      [?h :interaction.rearrangement/rearrangement ?rearrangement]
      [?int :interaction/rearrangement ?h]]
 
-    [(interaction->x ?int ?h ?gene)
+    [(interaction-> ?int ?h ?gene)
      [?int :interaction/interactor-overlapping-gene ?h]
      [?h :interaction.interactor-overlapping-gene/gene ?gene]]
-    [(interaction->x ?int ?h ?feature)
+    [(interaction-> ?int ?h ?feature)
      [?int :interaction/feature-interactor ?h]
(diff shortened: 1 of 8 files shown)


src/rest_api/classes/interaction/core.clj
```diff
@@ -45,16 +46,56 @@
...
+     [?h :interaction.feature-interactor/feature ?feature]
+     [?int :interaction/feature-interactor ?h]]
+    [(molecule->interaction-3 ?molecule ?h ?int)
+     [?h :interaction.molecule-interactor/molecule ?molecule]
+     [?int :interaction/molecule-interactor ?h]]
+    [(rearrangement->interaction-3 ?rearrangement ?h ?int)
+     [?h :interaction.rearrangement/rearrangement ?rearrangement]
+     [?int :interaction/rearrangement ?h]]
+    [(x->interaction-3 ?x ?h ?ix)
+     (or (gene->interaction-3 ?x ?h ?ix)
```
mgrbyte: a few questions and comments...

1.  New things I see are the introduction of "type" (e.g molecule, feature) specific interaction relations - were these being found before and are now just being literally expanded for readability?
 
2. I'm not entirely sure the naming of rule-heads (having the number-of-params as a suffix as a naming thing)

3.  _seems_ counter intuitive to have an `(or` statement  (since rules of the same name naturally *or* together; but this is a very complex area, so can understand why it's easier to break the  rules up. Related to `1.` am sure.
Perhaps naming the rules individually with defs and making `int-roles` a composite of them would help?
In that way one could still keep the same "rule head" names (would perhaps remove the need to

mgrbyte (commented):
In general, code looks good to me.

sibyl229:
Hi @mgrbyte I noticed a bug in my code after you reviewed it. It took me a long time to realize what was wrong. Finally fixed it and, PR is ready again.

To respond to your suggestions and questions:

1)  Yes, the molecule, feature, and rearrangement interactors have always been part of this table. It's just refactoring. "Other interactor" wasn't shown before, but Chris indicated that it should be.

2) The number of parameter in the rule name is removed :)

3) I didn't know that rule heads of the same name are `or` ed together. I kinda have those type-based rule heads in case a query is only interested in getting interactors of a particular type. But, so far in the code, we haven't encountered such a case. Let me know if you have any strong feeling about this. I can rename those if you want.

mgrbyte:
1 &  2 :+1: 
3.
I think what I'd do here is use the same name for all rule heads, and:
 * Separate each rule out into it's own `def`  (capability to use the rules to fetch type specific interactors   
 * Make anyother def for  all-rules, that aggregates all the individual rules (current usage).

So instead of `gene->interaction`, 'molecule->interaction` et al, rule name would just be `interaction`.
but because splitting them out into different `def`s means you could still use them interdependently.

sibyl229:
Thanks a lot @mgrbyte ! I updated the rule heads to be the same and removed the redundant `or` clause. When a use case for individual interactor-based rule appears, I will separate them into functions.

mgrbyte (approved):
I like this much better! :+1: 
Only change I'd make is the naming:

`x->interaction to`  `->interaction`
`interaction->x` to  `interaction->`

... but that's being picky! :wink:

sibyl229:
Thanks a lot @mgrbyte ! I will fix that before pulling it in. I imagine it's an idiomatic thing in Clojure

PR #177, opened 2018-06-12
Collaborators were not showing up in "Worked With" table because

azurebrd:
role-colour didn't exist, which didn't exist because collaborators
don't show in the graph.  But role-colour is being used as validator
for roles that should be allowed.


epaule:
thanks @azurebrd

PR #178, opened 2018-06-20
Feature/265 interaction

sibyl229:
Hi @mgrbyte minor refactoring and performance tweaking here, based on what data is used in the UI. Basically, `interactions` field is *primarily* used to render the table, and *sometimes* also include data to render the graph (when the size of the graph is small). `interaction_details` field is *only* used to render the graph. 
`showall` flag is removed, and replaced with `include_details` to indicate whether data for rendering the graph is included.
Thanks!

### DIFF
--- a/src/rest_api/classes/interaction.clj
+++ b/src/rest_api/classes/interaction.clj
@@ -12,4 +12,6 @@
    {:overview overview/widget
     :interactions interactions/widget
     :references references/widget
-    :external_links external-links/widget}})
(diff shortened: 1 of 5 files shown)


sibyl229 (commented):
I walked Adam through this, and he thinks it's fine. I'm gonna pull it in for now because the code freeze is getting close. And I'd like curators to be able to test this. If something needs to be changed here, I can do so later.

PR #183, opened 2018-12-17
Feature/variation sequence

adamjohnwright:
This pull request contains the variation molecular details widget. I have left comments in the code indicating which variations were used to test/compare with that which is on staging. 

The current plan is to get the curators to test this endpoint. I will fix anything that they find and then it will be rolled out into production. If we roll out WS268 before this widget is ready we will comment it out in production.


sibyl229:
Hi @a8wright everything else looks great! One small bug I noticed is that for the `amino_acid_change` field. It's returning `[null]` instead of `null`, which confuses the UI. Good tests cases are silent mutations like: WBVar01112111

PR #187, opened 2019-01-18
Adding protein sequence for nonsense and readthrough

adamjohnwright:
I have made it so that we are using mutant_start mutant_stop and wildtype_start wildtype_stop.

With this change, we are adding protein sequence to Readthrough and Nonsense mutations.

Here are some:
**Missense:** https://wormbase.org/species/c_elegans/variation/WBVar01684110#-58-10
**Readthrough:** https://wormbase.org/species/c_elegans/variation/WBVar00941611#-58-10
**Nonsense:** https://wormbase.org/species/c_elegans/variation/WBVar00466445#-58-10

I have also made it so that information on Missense and Nonsense show up in the location affect part of the predicted-cds table. This all should be checked by curators:)


sibyl229:
@a8wright  Looks good! I merged it in

PR #188, opened 2019-01-21
Fix/269 paper author

sibyl229:
relates to WormBase/website#6802

- differentiate author and person in pack-obj
- use :author/id as label when displaying author names for a paper


adamjohnwright:
I think I remember writing that logic. Or at least logic with making author names appear correctly. I am glad to see it simplified!

sibyl229:
Thanks @a8wright ! The requirement got simplified 😄

PR #189, opened 2019-01-25
Feature/feature molecular details

adamjohnwright:
This widget had been tested with two main Features: WBsf000519 and WBsf019129. The intention is to have this widget to be consistent with the variation molecular details widget. 

The first feature listed above (WBsf000519) is an example of a feature where we do not have the sequence. And the second feature (WBsf019129) is an example of a feature where we are to be displaying the sequence. For these features, I am returning the method. I think most of them have a method. This would likely be interesting to the user. 

I have included code to figure out the strand of the feature and am returning the strand in the JSON. I think it would be nice to make this clear to the user. I am not sure if it is always positive but either way it should be clear.

### DIFF
--- a/src/rest_api/classes/feature.clj
+++ b/src/rest_api/classes/feature.clj
@@ -14,6 +14,6 @@
    {:overview overview/widget
     :history history/widget
     :associations associations/widget
-    ;:molecular_details molecular-details/widget
+    :molecular_details molecular-details/widget
     :evidence evidence/widget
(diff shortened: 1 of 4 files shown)


sibyl229 (approved):

sibyl229:
Hi @a8wright please feel free to merge!

sibyl229:
related to WormBase/website#6871

PR #190, opened 2019-01-25
construct genotype string from disease model

sibyl229:
WormBase/Website#6026

### DIFF
--- a/src/rest_api/classes/gene/widgets/human_diseases.clj
+++ b/src/rest_api/classes/gene/widgets/human_diseases.clj
@@ -129,7 +129,11 @@
        :data (reduce (fn [result obj]
                        (assoc result (:label (pack-obj obj)) (pack-obj obj)))
                      strain-data
-                     (concat entities variation-genes))})))
+                     (concat entities variation-genes))
+       :entities (->> entities
+                      (cons strain)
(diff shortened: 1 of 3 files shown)


src/rest_api/classes/gene/widgets/human_diseases.clj
```diff
@@ -74,6 +75,70 @@
...
+                     (cons strain non-strain-entities)
+                     non-strain-entities)
+          entities-str (->> entities
+                            (map (fn [e]
+                                   (if (:variation/id e)
+                                     (let [gene (variation-gene-fn e)]
+                                       (format "%s(%s)" (:label (pack-obj gene)) (:label (pack-obj e))))
+                                     (:label (pack-obj e)))))
+                            (cons (when strain-str
+                                    (clojure.string/replace strain-str #"\.$" "")))
```
adamjohnwright, line 129: Usually, we have been importing [clojure.string :as str] at the top of the file and then doing str/replace in the file I am fine either way. I think we decided to do so early on so that we could find the dependencies easily at the top of each file. This dependency is kinda innocuous though.

adamjohnwright (approved):
everything looks great. I just had the one comment. You did a great job of writing code that is quite clear

PR #191, opened 2019-02-01
Feature/print sequences

adamjohnwright:
Effects sequence widgets for PCR Oligo, Pseudogene, Transcript, CDS

This will need to go through substantial testing buy curators and requires significant changes to the UI. 

@mgrbyte I thought you might be interested in this as I used a couple atom variables. I think it was cleanest to do so and seems to work!

### DIFF
--- a/src/rest_api/classes/cds.clj
+++ b/src/rest_api/classes/cds.clj
@@ -4,7 +4,7 @@
     [rest-api.classes.cds.widgets.overview :as overview]
     [rest-api.classes.cds.widgets.location :as location]
     [rest-api.classes.cds.widgets.feature :as feature]
-    ;[rest-api.classes.cds.widgets.sequences :as sequences]
+    [rest-api.classes.cds.widgets.sequences :as sequences]
     [rest-api.classes.cds.widgets.reagents :as reagents]
     [rest-api.classes.cds.widgets.references :as references]
     [rest-api.classes.gene.expression :as gene-expression]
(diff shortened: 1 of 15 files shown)


sibyl229 (approved):
Looks good to me. We might need to change the "positive/negative" strand to "+/-".

mgrbyte (commented):
Logic looks involved, kudos for working this out!

Generally i like the code, but with regard to the use of atoms specifically - I don't think they're the right tool for the job here  (unless i'm missed something, there isn't any shared state/concurrency involved?).

Was the choice of atoms for "program flow"/"processing" ?
If so, then i think there are alternatives/"better" choices.

e.g: The calculation of `sequence-features` seems to a series of cumulative string transformations, so perhaps  ->> with a few helper functions.

Calculation of `positive-features`:  probably use `loop`/`recur` (because of `next-feature-start`)

adamjohnwright:
@mgrbyte Thank you for taking the time to look through the code. I actually fully agree with the sentiment of what you are saying I tried a couple of ways to do this. As I believe the way it is being done with atoms is working I am going to merge this code.

That being said I did look into using loop/recur and thought it would likely be a better Clojure type way to do things. I think it would definitely make sure that there are no issues with race conditions. At this point, I just want to get this code over to Sibyl. Hopefully, I will get a chance to change this code over in the near future to use loop/recur. 

BTW thanks for saying it looks involved - that is very kind. I am sure your nameserver work is much more involved. I would love to someday look through your work in more detail.

PR #194, opened 2019-02-08
Feature/269 disease model

sibyl229:
WormBase/website#6044
WormBase/website#6763


sibyl229:
@mgrbyte since Adam is away. Could you take a look at this PR? The main change is to extract the code that generates the human diseases widget on gene page, and use it for DO term, variation, and strain page.

mgrbyte (approved):

PR #200, opened 2019-08-23
Refactor and modify hardcoded GO slim

sibyl229:


adamjohnwright:
I am not sure why you are keeping a bunch of GO terms that are commented out. you might want to state in the code what they are there for. Other than that looks clean and looks like a good use of memoizing :)

sibyl229:
@a8wright ! I have removed the commented out code. Thanks!

PR #211, opened 2020-04-16
add transcript mature product field 

sibyl229:
WormBase/website#7039


sibyl229:
@adamjohnwright Thanks for the quick pull!

PR #215, opened 2020-04-18
Self interaction

sibyl229:
WormBase/website#7547


adamjohnwright:
I read it over. I definitely should really dive deep into this code at some point and try to understand it. It definitely is more complex than I can understand but just looking it over. That being said I have read it over and I somewhat understand what the changes are doing. Looks good.

sibyl229:
@adamjohnwright thanks a lot for looking it over. I understand it's very difficult to look at.

PR #218, opened 2020-04-29
fix handling of non-directional interactions

sibyl229:
@adamjohnwright No need to make a production release for this by itself. Everything this affects is pre-cached. So I'm go to update the cache on my end.

I marked the branch as hotfix simply to indicate that it can be released to production, if you need to hotfix production for other things.


adamjohnwright (approved):

sibyl229:
@adamjohnwright Thanks!

PR #220, opened 2020-05-19
return counts for rnai not observing the phenotype is large 

sibyl229:
WormBase/website#7196


sibyl229:
@adamjohnwright Thanks!

PR #234, opened 2020-09-22
Add graphview endpoints

adamjohnwright:
This took a bit more (twice) the work I was expecting. The main difficulty was getting it to work with all the ways the data is arranged in the database. I think I have got it now. Let me know if you run into an exception. 

Also, I am calling these endpoints "graphview". I am open to suggestions. 

In this pull request, we are adding endpoints to see data related to the page entity within one hop away. These endpoints provide the data in both the "forward" and "backward" direction from the entity in question. It also provides schema information that is provided by the pseudo ace package. In this schema data, there is some information on how the data is related to the AceDB data and if it is being converted. I think it would be nice to have this information more readily available to both developers and curators. 

The code has been written to use pack-obj on entities related to the one in question that has an associated page. It has also been written to only display the count if there are more than 30 entities in a relationship. This number could be adjusted based on performance. 

I think that these endpoints are very useful with or without an associated widget. 

If we are to create a widget I am starting to think that an expandable tree view would be a great way of displaying the data. I am thinking something like one of these: https://www.cssscript.com/tag/tree-view/

The main complicated part would be making the external links from the pack-obj(s).


sibyl229:
Hi @adamjohnwright I'm okay with you pulling this in, if this helps you, and you are confident that this won't affect the performance of the production REST API negatively. I took a look at your code, but did not do any benchmarking. 

The discussion about the UI would have to happen separately, to minimize duplicated development effort, and to ensure the effort produce something that internal members would actually use. For a starter, things that might need to be done first are: a closer examination of what is offered by TrACeViewer (https://github.com/WormBase/down. It can be run similar to wormbase_rest. I tried it last week), and validation of the design with the intended users with respect to the tasks the tool intends to facilitate. Also, more thought might need to go in on what edges cases there are, how they are handled and their impact, for example, reducing the sets with size larger than 30 to a count might limit the usefulness of the tree view to curators.

adamjohnwright:
@sibyl229 thanks. I agree with your assessment. I think that sometime soon we should figure out who we should talk with to figure out what would be most useful for curators.

PR #235, opened 2020-09-30
278 Add genotype class and update disease association to use genotype

sibyl229:


adamjohnwright (approved):
All looks good

sibyl229:
@adamjohnwright great! Thanks for checking

PR #238, opened 2021-01-05
fix interactor text format to be consistent with other interactor types

sibyl229:
Can be released anytime. WormBase/website#7950


adamjohnwright (approved):
Looks good. I am not sure if it will ever fail. You might want to make sure it is not null before passing it to pack-obj. I forget exactly when or if you need to check

sibyl229:
I checked that `pack-obj` is handling null as expected.

PR #242, opened 2021-04-01
vep related molecular-changes for transcript

sibyl229:
WormBase/website#8151

@adamjohnwright for start, I made the changes based on @markquintontulloch 's mockup attached to the issue above. I expect these changes to only affect the transcripts. Please let me know if there is something I need to change.

### DIFF
--- a/src/rest_api/classes/variation/widgets/molecular_details.clj
+++ b/src/rest_api/classes/variation/widgets/molecular_details.clj
@@ -681,6 +681,78 @@
                                    {:item
                                     (pack-obj t)
 
+                                    :vep_consequence
+                                    (some-> (:molecular-change/vep-consequence h)
+                                            (first)
+                                            (str/replace "_" " "))
+
+                                    :vep_impact
+                                    (some->> (:molecular-change/vep-impact h)
+                                             (first)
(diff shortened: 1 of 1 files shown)


src/rest_api/classes/variation/widgets/molecular_details.clj
```diff
@@ -681,6 +681,78 @@
                                    {:item
                                     (pack-obj t)
 
+                                    :vep_consequence
+                                    (some-> (:molecular-change/vep-consequence h)
+                                            (first)
+                                            (str/replace "_" " "))
```
adamjohnwright, line 687: I think this might be an error if the stirring is nil. I am not sure if you need to do a check first. same with the rest of the fields added in this pull request. If not everything looks great

adamjohnwright (approved):

adamjohnwright (approved):

src/rest_api/classes/variation/widgets/molecular_details.clj
sibyl229: Thanks for checking! `some->` should short circuit the thread if `nil` is returned by any step

PR #245, opened 2021-05-13
lining up predicted cds with cds and amino acid 

adamjohnwright:
This change affects all pages where we have the protein sequence for the variation here is the main one that I found: https://staging.wormbase.org/species/c_elegans/variation/WBVar00466445#69--10

Fortunately, most of the logic is the exact same as before for generated the AA sequences.


sibyl229 (approved):
Looks good to me!

PR #246, opened 2021-05-26
fixing genetics widget and bug fix for molecular details widget

adamjohnwright:
I have this running on staging right now

@Paul-Davis is going to be looking into whether or not it needs changes. @sibyl229 This PR is mainly to check the code seems correct. In any case, I think that populating the columns with this data is a lot better than what is in production.

### DIFF
--- a/src/rest_api/classes/variation/core.clj
+++ b/src/rest_api/classes/variation/core.clj
@@ -42,6 +42,10 @@
        (set)))
 
 
+(defn remove-when-all-nil [x]
+ (when (some #(not (nil? %)) x) x))
+
+
 (defn process-variation [var relevant-location?
                          & {:keys [window]
                             :or {window 20}}]
@@ -130,7 +134,8 @@
                        "5' UTR")
                      (when (:molecular-change/three-prime-utr tc)
                        "3' UTR"))]
-                    (str/join "," effects))))
+                    (str/join ", " effects))))
+              (remove-when-all-nil)
               (not-empty))
 
      :effects
@@ -138,6 +143,7 @@
        (->> changes
             (map molecular-change-effects)
             (filter identity)
+            (remove-when-all-nil)
             (not-empty)))
 
      :composite_change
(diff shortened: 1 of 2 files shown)


sibyl229 (commented):
The use of `(distinct)` and `(remove nil?)` might make the logic incorrect. 

The reason is that the **first** item in the every column (effect, composite_change, location) should correspond to the **first** CDS. The second item corresponds to the second CDS. They are essentially **sub-rows within a row**, where a row is about a variation, and a subrow is about a CDS.

Once `(distinct)` and `(remove nil?)` is called, you can no longer tell what corresponds to what.

In addition, the use of `str/join` is to preserve the correspondence within a sub-row (about a CDS). In other words, you don't need to `str/join` the result of **between different CDSs (different sub-rows)**, returning them as an array is sufficient, which will put them on different sub -rows as intended. You only need `(str/join ', ')`, if those values belong to the **same CDS (same sub-row)**.

adamjohnwright:
@sibyl229 thanks for putting thought into this. I had an inclination in the back of my mind that this might not be correct but I didn't have any deeper insight into it. I will look into addressing your thoughts and will contact you when I run into questions.

adamjohnwright:
@sibyl229 I ended up getting sick (not covid - thank God). I am now on the other side of the cold. I spend a few hours here trying to reason things through. I have made it so that there is an extra field for the transcript. I think we need to add a column for it in some circumstances. I have made it so that the nils are not removed. This makes it so that bullet points are there and empty quite often. I am not sure what we do about them because they are useful in lining things up in rare (but not too rare circumstances). I feel like at, this point, it is a matter of adjusting the display. I also would change 'isotope' to 'CDS' as I am not providing links to the proteins. 

Here are a few links that I have been using to look at the results:

https://staging.wormbase.org/species/c_elegans/gene/WBGene00003055#-4f-10
https://staging.wormbase.org/species/c_elegans/gene/WBGene00000101#-4f-10
https://staging.wormbase.org/species/c_elegans/gene/WBGene00015146#-4-10

Please let me know if you see any issues or have any further ideas of how things should line up.

PR #247, opened 2022-05-25
BioCyc pathways endpoint was added

nuin:
JSON file with BioCyc pathways was added to resources, need to define logistics about it


adamjohnwright (approved):

nuin:
thanks @adamjohnwright

